Issue #8, opened 2018-02-06
Changes required on each apply


cliveza:
Hi - This module is great, but each time I run terraform apply it looks like it wants to set and then remove a certificate - just wondering if you know why?

```viewer_certificate.111111111.acm_certificate_arn:            "" => "arn:aws:acm:us-east-1:12121212:certificate/1111111-222222-3333-4444-4444"
viewer_certificate.111111111.cloudfront_default_certificate: "" => "true"
viewer_certificate.111111111.iam_certificate_id:             "" => ""
viewer_certificate.111111111.minimum_protocol_version:       "" => "TLSv1"
viewer_certificate.111111111.ssl_support_method:             "" => "sni-only"
viewer_certificate.222222222.acm_certificate_arn:             "arn:aws:acm:us-east-1:12121212:certificate/1111111-222222-3333-4444-4444" => ""
viewer_certificate.222222222.cloudfront_default_certificate:  "false" => "false"
viewer_certificate.222222222.iam_certificate_id:              "" => ""
viewer_certificate.222222222.minimum_protocol_version:        "TLSv1" => ""
viewer_certificate.222222222.ssl_support_method:              "sni-only" => ""```

**Comment by osterman** [comment]
In my experience, I've seen this happen if the format of your ARN doesn't match the normalized version that AWS or Terraform uses. Can you share the complete plan?

**Comment by cliveza** [comment]
I have a website module that I use to create a prod, staging and dev version of the a static website. 

This website module calls the terraform-aws-cloudfront-s3-cdn module 3 times, once for each environment.

**website.tf file:**
```
data "aws_acm_certificate" "my_domain_com" {
  domain   = "*.my_domain.com"
  statuses = ["ISSUED"]
}

module "frog_website" {
  source              = "../modules/website"
  subdomain           = "frog"
  domain              = "my_domain.com"
  acm_certificate_arn = "${data.aws_acm_certificate.my_domain_com.arn}"
}
```

**website module:**
```
locals {
  name         = "${var.subdomain}.${var.domain}"
  prod_fqdn    = "${local.name}"
  staging_fqdn = "${var.subdomain}-staging.${var.domain}"
  dev_fqdn     = "${var.subdomain}-test.${var.domain}"
}

module "development_website" {
  source              = "git::https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn.git?ref=master"
  namespace           = "my-name-space"
  stage               = "dev"
[...]

PR #9, opened 2018-02-12
Added conditional

cliveza:
## what
* Set `cloudfront_default_certificate`

## why
* Must specify acm_certificate_arn, or iam_certificate_id - but not both
* https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/issues/8


main.tf
```diff
@@ -128,7 +128,7 @@ resource "aws_cloudfront_distribution" "default" {
     acm_certificate_arn            = "${var.acm_certificate_arn}"
     ssl_support_method             = "sni-only"
     minimum_protocol_version       = "TLSv1"
-    cloudfront_default_certificate = true
+    cloudfront_default_certificate = "${var.acm_certificate_arn == "" ? 1 : 0}"
```
aknysh: nice catch @cliveza 

since `cloudfront_default_certificate` is boolean, should we change it to 
```hcl
cloudfront_default_certificate = "${var.acm_certificate_arn == "" ? true : false}
```

instead of `1 : 0` (although this would work too)

PR #19, opened 2018-07-04
Migrate to README.yaml format

vadim-hleif:
## what
* Add `README.yaml`

## why
* Standardize README


outputs.tf
```diff
@@ -1,31 +1,39 @@
 output "cf_id" {
-  value = "${aws_cloudfront_distribution.default.id}"
+  value       = "${aws_cloudfront_distribution.default.id}"
+  description = "ID of AWS CloudFront distribution"
 }
 
 output "cf_arn" {
-  value = "${aws_cloudfront_distribution.default.arn}"
+  value       = "${aws_cloudfront_distribution.default.arn}"
+  description = "ID of AWS CloudFront distribution"
```
sarkis: ARN of AWS CloudFront distribution

variables.tf
```diff
@@ -32,7 +32,8 @@ variable "tags" {
 }
 
 variable "enabled" {
-  default = "true"
+  default     = "true"
+  description = "State of CloudFront"
```
sarkis: Boolean which determines if CloudFront resources are created

variables.tf
```diff
@@ -110,39 +120,47 @@ variable "log_expiration_days" {
...
+  default     = "false"
+  description = "Forward query strings to the origin that is associated with this cache behavior"
 }
 
 variable "cors_allowed_headers" {
-  type    = "list"
-  default = ["*"]
+  type        = "list"
+  default     = ["*"]
+  description = "List of allowed headers  for S3 bucket"
```
sarkis: unneeded extra space between words `headers` and `for`

variables.tf
```diff
@@ -110,39 +120,47 @@ variable "log_expiration_days" {
...
 }
 
 variable "cors_allowed_methods" {
-  type    = "list"
-  default = ["GET"]
+  type        = "list"
+  default     = ["GET"]
+  description = "List of allowed methods (e.g. ` GET, PUT, POST, DELETE, HEAD`) for S3 bucket"
```
sarkis: Do we need the backtick here? If so maybe we should do e.g. `GET`, `PUT`, ...

variables.tf
```diff
@@ -110,39 +120,47 @@ variable "log_expiration_days" {
...
 }
 
 variable "cors_allowed_origins" {
-  type    = "list"
-  default = []
+  type        = "list"
+  default     = []
+  description = "List of allowed origins (e.g. ` example.com, test.com`) for S3 bucket"
```
sarkis: do we need the ` here?

variables.tf
```diff
@@ -151,45 +169,54 @@ variable "viewer_protocol_policy" {
 }
 
 variable "allowed_methods" {
-  type    = "list"
-  default = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
+  type        = "list"
+  default     = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
+  description = "List of allowed methods (e.g. ` GET, PUT, POST, DELETE, HEAD`) for AWS CloudFront"
```
sarkis: do we need the ` here?

variables.tf
```diff
@@ -151,45 +169,54 @@ variable "viewer_protocol_policy" {
...
 }
 
 variable "cached_methods" {
-  type    = "list"
-  default = ["GET", "HEAD"]
+  type        = "list"
+  default     = ["GET", "HEAD"]
+  description = "List of cached methods (e.g. ` GET, PUT, POST, DELETE, HEAD`)"
```
sarkis: do we need the ` here?

variables.tf
osterman: yea, let's just remove the back tick

PR #20, opened 2018-07-18
make minimim_protocol_version configurable

jschaul:
Make minimim_protocol_version configurable.

(I wasn't able to run `make readme` to completion as this repo and/or the included build harness fails due to presupposed binaries (`/bin/bash: gomplate: command not found`) whose origins are not documented)

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,7 @@ Available targets:
 | log_standard_transition_days | Number of days to persist in the standard storage tier before moving to the glacier tier | string | `30` | no |
 | max_ttl | Maximum amount of time (in seconds) that an object is in a CloudFront cache | string | `31536000` | no |
 | min_ttl | Minimum amount of time that you want objects to stay in CloudFront caches | string | `0` | no |
+| minimum_protocol_version | Cloudfront TLS minimum protocol version | string | `TLSv1` | no |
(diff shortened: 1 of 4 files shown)


osterman:
@jschaul sorry about the trouble.

Run `make readme/deps` to download gomplate. I'll see about making that process smoother.

osterman:
@jschaul, sorry we missed your PR. Join our slack team and you can reach us directly in #terraform channel.

[Slack Community](https://slack.cloudposse.com)

aknysh:
@jschaul  looks good, thanks.
please run `make readme/deps` and then `make readme` to update `README`.

jschaul:
Updated and rebased. Should be good to be merged now.

(Your readme regeneration process is slightly clunky; I had to manually commit only parts of the generated readme, since that build harness doesn't update automatically, in order to avoid undoing the changes introduced by #21)

osterman:
@jschaul thanks for the feedback. Always looking to improve the process. There should be zero need to commit certain parts (haven't run into that before). If you can share screenshots, perhaps it will help me understand the problem.

To update the build-harness, just re-run `make init`

Open to suggestions. Don't think we automatically want to update the build-harness. 

 We recently went through the process of updating all ~120 modules with our `README.yaml` spec.

PR #25, opened 2018-08-27
Forwarded headers are needed when using CORS on S3 from cloudfront

Jamie-BitFlight:
## what:
* Add forwarded headers so that s3 bucket can identify origins.

## why
* Its missing and causes an error

## references
* https://aws.amazon.com/premiumsupport/knowledge-center/no-access-control-allow-origin-error/

### DIFF
--- a/docs/terraform.md
+++ b/docs/terraform.md
@@ -21,6 +21,7 @@
 | delimiter | Delimiter to be used between `name`, `namespace`, `stage`, etc. | string | `-` | no |
 | enabled | Select Enabled if you want CloudFront to begin processing requests as soon as the distribution is created, or select Disabled if you do not want CloudFront to begin processing requests after the distribution is created. | string | `true` | no |
 | forward_cookies | Time in seconds that browser can cache the response for S3 bucket | string | `none` | no |
+| forward_header_values | A list of whitelisted header values to forward to the origin | list | `<list>` | no |
 | forward_query_string | Forward query strings to the origin that is associated with this cache behavior | string | `false` | no |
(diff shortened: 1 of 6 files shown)


main.tf
```diff
@@ -144,6 +144,7 @@ resource "aws_cloudfront_distribution" "default" {
 
     forwarded_values {
       query_string = "${var.forward_query_string}"
+      headers      = ["Access-Control-Request-Headers", "Access-Control-Request-Method", "Origin"]
```
osterman: Should we make this a `variable` with a default of `["Access-Control-Request-Headers", "Access-Control-Request-Method", "Origin"]`?

aknysh: Although they mention only the three headers for S3 origin

```
If your origin is an S3 bucket, you typically must configure your distribution to forward and whitelist the following headers to Amazon S3:

Access-Control-Request-Headers
Access-Control-Request-Method
Origin
```

It's a good idea to move it to a variable so it could be changed if needed.

Jamie-BitFlight:
Moved, and regen'd the readme.

docs/terraform.md
```diff
@@ -21,6 +21,7 @@
 | delimiter | Delimiter to be used between `name`, `namespace`, `stage`, etc. | string | `-` | no |
 | enabled | Select Enabled if you want CloudFront to begin processing requests as soon as the distribution is created, or select Disabled if you do not want CloudFront to begin processing requests after the distribution is created. | string | `true` | no |
 | forward_cookies | Time in seconds that browser can cache the response for S3 bucket | string | `none` | no |
+| forward_header_values | A list of whitelisted header values to forward to the origin | list | `<list>` | no |
```
osterman: nit: `forward_header_values` is not consistently named relative to the other parameters in this group. Should simply be `forward_headers`

PR #38, opened 2019-01-28
Add dns_enable to variables

ommadawn46:
## Overview
I add a dns_enable option to variables in this PR.
Thanks to this option, it's possible to build only CloudFront and S3.

## Purpose
Current Route53 setting only creates A record. This PR makes it possible to use customized Route53 module: such as using other DNS record types, setting failover, handling health check, and so on. 
To do that, I add the option to create only CloudFront and S3.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -173,7 +173,7 @@ resource "aws_cloudfront_distribution" "default" {
 
 module "dns" {
   source           = "git::https://github.com/cloudposse/terraform-aws-route53-alias.git?ref=tags/0.2.5"
-  enabled          = "${var.enabled == "true" && var.dns_enabled == "true" ? "true" : "false" }"
+  enabled          = "${var.enabled == "true" && var.dns_enabled == "true" ? "true" : "false"}"
   aliases          = "${var.aliases}"
   parent_zone_id   = "${var.parent_zone_id}"
   parent_zone_name = "${var.parent_zone_name}"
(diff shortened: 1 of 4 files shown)


main.tf
```diff
@@ -173,7 +173,7 @@ resource "aws_cloudfront_distribution" "default" {
 
 module "dns" {
   source           = "git::https://github.com/cloudposse/terraform-aws-route53-alias.git?ref=tags/0.2.5"
-  enabled          = "${var.enabled}"
+  enabled          = "${var.enabled ? var.dns_enabled : false}"
```
joshmyers: `enabled          = "${var.enabled == "true" && var.dns_enabled == "true" ? "true" : "false" }"`

ommadawn46: @joshmyers 
Thank you for review. I fixed it.

ommadawn46:
@joshmyers 
Would you please review again?

variables.tf
```diff
@@ -275,3 +275,8 @@ variable "web_acl_id" {
   default     = ""
   description = "ID of the AWS WAF web ACL that is associated with the distribution"
 }
+
+variable "dns_enabled" {
+  default     = "true"
+  description = "Select Disabled if you do not want to create DNS config after the distribution is created"
```
```suggestion
  description = "Set to `false` if you do not want to create the DNS hostname after the distribution is created"
```

ommadawn46:
@aknysh Thank you for the review.

I changed to the same way as terraform-aws-rds-cluster, and deleted dns_record.
Without the following commit, it did not work, so I updated the module version.
https://github.com/cloudposse/terraform-aws-route53-alias/commit/74957ad27314622bc1d647f98216a1c8d6ee8130

Issue #50, opened 2019-09-20
Error putting S3 policy: MalformedPolicy: Invalid principal in policy


hemanthudimella:
I guess this is a blocker.
This module wouldn't work unless AWS fixes this issue or you start using canonical ID instead of CloudFront Origin Access Identity ID.
Recently AWS made a few changes to their origin access identity policy format. The S3 bucket policy is now rejecting principal with spaces in it.
"Principal": {
  "AWS": "arn:aws:iam::cloudfront:user/**CloudFront Origin Access Identity EO3XXXXXXXX**"
 }
They now replaced it with  
"Principal": {
  "AWS": "arn:aws:iam::cloudfront:user/**CloudFront_Origin_Access_Identity_EO3XXXXXXXX**"
 }
This module is still generating policy with spaces in its principal and hence the issue.
Can we change it to use canonical ID instead?

gyoza:
I am having this issue too and I thought I was going crazy!!!

gyoza:
`# module.computercompany_cdn_dev.aws_s3_bucket_policy.default will be created   
  + resource "aws_s3_bucket_policy" "default" {
      + bucket = "dev-computercompanyccc-net-origin"
      + id     = (known after apply)
      + policy = jsonencode(
            {
              + Statement = [
                  + {
                      + Action    = "s3:GetObject"
                      + Effect    = "Allow"
                      + Principal = {
                          + AWS = "arn:aws:iam::cloudfront:user/CloudFront Origin Access Identity "
                        }
                      + Resource  = "arn:aws:s3:::dev-computercompanyccc-net-origin/*"
                      + Sid       = ""
                    },
                  + {
                      + Action    = "s3:ListBucket"
                      + Effect    = "Allow"
                      + Principal = {
                          + AWS = "arn:aws:iam::cloudfront:user/CloudFront Origin Access Identity "
                        }
                      + Resource  = "arn:aws:s3:::dev-computercompanyccc-net-origin"
                      + Sid       = ""
                    },
                ]
              + Version   = "2012-10-17"
            }
        )
    }`

I think its a problem with the labeler.

this is in the tfstate

[...]

PR #51, opened 2019-09-20
Use canonical user Id instead of iam user to fix s3 invalid policy issue

hemanthudimella:
This PR fixes https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/issues/50

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,13 +308,13 @@ Check out [our other projects][github], [follow us on twitter][twitter], [apply 
 |---|---|---|---|
 
   [osterman_homepage]: https://github.com/osterman
-  [osterman_avatar]: https://github.com/osterman.png?size=150
+  [osterman_avatar]: https://img.cloudposse.com/150x150/https://github.com/osterman.png
   [aknysh_homepage]: https://github.com/aknysh
-  [aknysh_avatar]: https://github.com/aknysh.png?size=150
+  [aknysh_avatar]: https://img.cloudposse.com/150x150/https://github.com/aknysh.png
   [Jamie-BitFlight_homepage]: https://github.com/Jamie-BitFlight
-  [Jamie-BitFlight_avatar]: https://github.com/Jamie-BitFlight.png?size=150
+  [Jamie-BitFlight_avatar]: https://img.cloudposse.com/150x150/https://github.com/Jamie-BitFlight.png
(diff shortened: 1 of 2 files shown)


gyoza:
This will cause your applies to change the policy on every run.

hemanthudimella:
Done. Rebuilt README.

aknysh:
@hemanthudimella 
please see comments from @gyoza  https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/51#issuecomment-533594653

you just need to replace ” “, “_” on the old value

hemanthudimella:
@gyoza @aknysh 
Did you mean replace manually? That means every time I apply a change, terraform removes my manual change, and I will have to do it again.
And if you didn't mean manual, please note that replacing " " with "_" works for all new policies. But any policies that are existing wouldn't accept "_" and `terraform apply` fails. (Existing policies are still accepting " " and " " only)
From my chat with  AWS support, the recommended way is to use canonical user id instead.
They were also mentioning that the syntax is not yet finalized and we can expect the "_" to be replaced or removed in future without any notice as they replaced " " with "_" without notice.
However I agree that `terraform plan` will always show a change in the policy, if we use canonical user id (even though there is no change in the policy). Hopefully terraform fixes this soon.
But until then using canonical user id is the only way (at least as far as i know) to make both existing and new policies work.

gyoza:
Just commenting on what I found when I encountered this issue.

 Our CICD process does "things" when terraform reports items to change. So
when this is happening when there are actually no changes is somewhat a big
deal.

This is probably a problem for more than just our process.

For now I've opted to fork this module and hardcode appropriate changes
until the provider is fixed.

hemanthudimella:
@gyoza Same here. I ended up forking the module. :)
@aknysh I suppose this module, at the moment, is unusable as is.

aknysh:
@hemanthudimella 
please run `terraform fmt` (our CI/CD tests are failing).
After that, we'll merge the PR

aknysh:
@hemanthudimella  please rebase and run `terraform fmt` and then rebuild README:

make init
make readme/deps
make readme
```

maximmi:
@hemanthudimella is this PR still actual due to changes made and merged after this one?

hemanthudimella:
Redundant PR

PR #52, opened 2019-10-30
additional bucket policy

balihb:
Solution in case there are additional policies needed for the bucket.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -66,7 +66,7 @@ variable "use_regional_s3_endpoint" {
 variable "additional_bucket_policy" {
   type        = string
   default     = "{}"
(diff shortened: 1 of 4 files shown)


variables.tf
```diff
@@ -63,6 +63,12 @@ variable "use_regional_s3_endpoint" {
   default     = false
 }
 
+variable "additional_bucket_policy" {
+  type        = string
+  default     = "{}"
+  description = "Additional policies for the bucket. The variables $${bucket_name} and $${origin_path} will be substituted."
```
```suggestion
  description = "Additional policies for the bucket. If included in the policies, the variables $${bucket_name} and $${origin_path} will be substituted."
```

Issue #57, opened 2020-01-08
Allow removing `origin` as an attribute on the origin label


dmattia:
As best I can tell, this restricts the naming of the s3 origin bucket to something ending in "origin".

I have no problem with this in theory, except for the fact that I'm migrating an existing terraform setup I wrote a ways back to try to use this module instead. The existing bucket I have as the origin of my current infra does not end in "origin".

Forcing me to change the name of my s3 bucket is a significant annoyance, as CI systems need to be updated, content needs to be transferred, etc. which can cause downtime.

Having a var like `var.extra_origin_bucket_attributes` with a default of `["origin"]` would be wonderful.

PR #58, opened 2020-01-08
Allow removing `origin` from the origin label

dmattia:
I also went ahead and did the same with "logs" on the log bucket.

These changes are backwards compatible, and would close https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/issues/57


dmattia:
Thanks for the review @aknysh, I rebuilt the README.md and formatted the code

aknysh:
@dmattia 
thanks again!

All is good except the Terratest timeouts

```
test timed out after 30m0s
```


Here https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/blob/master/test/src/Makefile#L45

Can you please update it to 60m timeout:

```
cd $(BASE) && go test -v -timeout 60m -run TestExamplesComplete
```

dmattia:
Updated from 30m to 60m

Issue #56, opened 2020-01-08
Add `aliases` to outputs


dmattia:
In the CORS settings of some other modules, it would be nice to easily get all aliases that a CF dist made through this module controls.

Something like:

```terraform
output "aliases" {
  value = var.aliases
}
```

would be wonderful.

PR #59, opened 2020-01-08
Added `aliases` to outputs

dmattia:
In the CORS settings of some other modules, it would be nice to easily get all aliases that a CF dist made through this module controls.

Would fix: https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/issues/56


dmattia:
Thanks for the review @aknysh, I rebuilt the README.md and formatted the code

Issue #62, opened 2020-01-09
Add support for `ordered_cache_behavior`


dmattia:
I would like to handle the caching of some files differently from others.

In my particular case, this means handling `*.html` paths to not be cached at all.

My favorite option would be something like:

```hcl
variable "caching_blacklist" {
  type = set(string)
  default = []
  description = "Paths of objects that should never be cached for any HTTP methods"
}
```

where I could specify `caching_blacklist = ["*.html"]` in my vars, but more cases would be covered by covering all ordered_cache_behavior settings, with something like:

```hcl
variable "ordered_cache_behavior" {
  type = list(object({
    path_pattern = string
    allowed_methods = string
    ...
  }))
  default = []
  description = "..."
}
```

CarpathianUA:
+1

PR #63, opened 2020-01-09
Added support for blacklisting paths for caching

dmattia:
This is related to https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/issues/62


aknysh:
@dmattia  thanks for the PR.
I think it would be better to provide an `cache_behaviors` var and then contract `ordered_cache_behavior` block from it.
Otherwise we are using that block just for blacklisting, which applies to a limited set of use-ceses.

dmattia:
Makes sense. I wasn't actually sure if for blocks that allowed repeats like `ordered_cache_behavior` if you could have multiple separate `dynamic` loops to generate them from different sources.

I can make this change, but it may take a while to prioritize it on my end. If anyone else wants to pick it up, please feel free.

dmattia:
I can make that change! I'd love to talk about your opinions on the var types though.

The `default_cache` and `ordered_cache`s are almost entirely the same, except the default cache doesn't have a path option (at least that's my understanding).

So my thought was that we would leave the existing cache vars as they are, updating their docs to specify that they are only for the default cache. And then the ordered_caches would have a `var.ordered_cache_behaviors` where clients would have to pass in a large array of object with fields related to caching stuffs.

So basically what you said, but with `var.ordered_cache_behaviors` instead of `var.cache_behaviors`. Does that sound okay?

dmattia:
I created https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/70 to more generally support ordered caches

dmattia:
I believe it would be possible to support both PRs, but will leave it up to you guys to determine if you think that would be too much

maximmi:
@dmattia let's stick to more general way! thanks for your contribution

PR #67, opened 2020-01-15
Make it optional to override the origin s3 policy

dmattia:
When specifying var.origin_bucket, it can be nice to leave the existing bucket's policy as is.

As an example, I manage an s3 bucket that multiple CloudFront dists use as their origin (without paths, they just use the same code).

In this case, I do not want to restrict the bucket to only talk to a single CF dist, as this module does by default.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -69,6 +69,7 @@ data "template_file" "default" {
 }
 
 resource "aws_s3_bucket_policy" "default" {
+  count  = ! local.using_existing_origin || var.override_origin_bucket_policy ? 1 : 0
   bucket = local.bucket
   policy = data.template_file.default.rendered
 }
@@ -77,7 +78,7 @@ data "aws_region" "current" {
 }
 
 resource "aws_s3_bucket" "origin" {
-  count         = signum(length(var.origin_bucket)) == 1 ? 0 : 1
+  count         = local.using_existing_origin ? 0 : 1
   bucket        = module.origin_label.id
   acl           = "private"
   tags          = module.origin_label.tags
@@ -148,6 +149,8 @@ data "aws_s3_bucket" "selected" {
 }
 
 locals {
+  using_existing_origin = signum(length(var.origin_bucket)) == 1
+
   bucket = join("",
     compact(
       concat([var.origin_bucket], concat([""], aws_s3_bucket.origin.*.id))
--- a/variables.tf
+++ b/variables.tf
@@ -81,6 +81,12 @@ variable "additional_bucket_policy" {
   description = "Additional policies for the bucket. If included in the policies, the variables `$${bucket_name}`, `$${origin_path}` and `$${cloudfront_origin_access_identity_iam_arn}` will be substituted. It is also possible to override the default policy statements by providing statements with `S3GetObjectForCloudFront` and `S3ListBucketForCloudFront` sid."
 }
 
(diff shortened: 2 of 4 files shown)


aknysh:
@dmattia please rebase

dmattia:
Just rebased @aknysh

variables.tf
```diff
@@ -81,6 +81,12 @@ variable "additional_bucket_policy" {
   description = "Additional policies for the bucket. If included in the policies, the variables `$${bucket_name}`, `$${origin_path}` and `$${cloudfront_origin_access_identity_iam_arn}` will be substituted. It is also possible to override the default policy statements by providing statements with `S3GetObjectForCloudFront` and `S3ListBucketForCloudFront` sid."
 }
 
+variable "override_origin_bucket_policy" {
+  type        = string
```
```suggestion
  type        = bool
```

dmattia:
Oops, thanks for catching that! Updated and rebased @aknysh

dmattia:
yep, I have modules using this code already, both with and without an existing origin given and have had no problems

main.tf
```diff
@@ -69,6 +69,7 @@ data "template_file" "default" {
 }
 
 resource "aws_s3_bucket_policy" "default" {
+  count  = local.using_existing_origin && ! var.override_origin_bucket_policy ? 0 : 1
```
aknysh: @dmattia 
do we really need `var.override_origin_bucket_policy`?
If you specify `var.origin_bucket`, the policy will not be created as well as the bucket itself.
If you don't specify the origin bucket, we want to create it as well as the policy.

aknysh:
@dmattia  please rebase and address the comments, thanks again

main.tf
dmattia: I would prefer the approach you're talking about here, where we only manage the policy if we manage the bucket.

The reason I added the var is for backwards compatibility, as the current code in master _will_ manage the policy even if the bucket isn't managed.

If you're okay with removing var.override_origin_bucket_policy, I'll gladly do it!

@aknysh

aknysh:
```
count  = local.using_existing_origin && ! var.override_origin_bucket_policy ? 0 : 1
```

In the expression above, if the existing bucket is provided, the policy will not be created because `local.using_existing_origin` is `true`, so `! var.override_origin_bucket_policy` is completely redundant 

Please update the logic.
So yes, we want to manage the policy if we create the bucket.
And make it configurable to manage the policy or not when an existing bucket is provided.
(this is not the expression above does. It will never create the policy if an existing bucket is provided).



Thanks

dmattia: (Apologies for the delay, I was OOO).

Thanks for the correction! I updated the logic, not sure how I made that mistake the first time round

dmattia:
@aknysh rebased and updated code

dmattia:
Ran `fmt`, and fixed the var name so that `validate` passes

maximmi:
@dmattia can't make a fix because of `permission denied` to push changes to your fork
```
! [remote rejected] dmattia/s3_policy -> dmattia/s3_policy (permission denied)
error: failed to push some refs to 'git@github.com:dmattia/terraform-aws-cloudfront-s3-cdn.git'
```
could you please check for permissions there?

dmattia:
@maximmi I added you as a contributor to the repo

Issue #14, opened 2018-02-26
Configure S3 to support `website` mode


mojeto:
Hi,
How can I use your module to configure S3 origin as `Static website hosting` ?
It's disabled by default. 
The main reason for me is `Static website hosting` allows mapping 'folder' and 'subfolder' url paths to `index.html` file in given folder. 
i.e  `/contact/`  -> `/contact/index.html`

With current S3 setup I get `/contact/` Error `NoSuchKey`.

osterman:
As implemented, it might not support that. Our client's use-case was to specifically to serve static assets. The module could probably be extended to add that support.

In the meantime, you can achieve the same thing by using our 2 other modules:
- https://github.com/cloudposse/terraform-aws-s3-website
- https://github.com/cloudposse/terraform-aws-cloudfront-cdn

Basically, create the S3 "website" bucket and then accelerate with our generic `cloudfront-cdn` module which will work with any website.

mojeto:
Thanks for a good work you have done on those modules and for pointing me to correct direction too. 

I leave this issue open in a case you want to implement this feature in future, but feel free to close it as I'll use the other two modules you suggested.

Cheers.

osterman:
thanks @mojeto! I'll keep it open for now. Makes sense that we should add support for `website` mode.

geota:
I tried combining the above modules and didnt find it satisfactory for a couple of reasons:

1) s3-website only allows for acl "public-read", id like to avoid this and serve the assets only through the CDN using aws_cloudfront_origin_access_identity resource like you do currently.
2) there is no ability to inject your own buckets which then leads to two log buckets being created (one per each module)

Im going to attempt to submit a PR here for this.

aknysh:
addressed in https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/releases/tag/0.15.0

PR #73, opened 2020-01-23
Add website mode to use S3 static website as origin

othree:
For some reason https://github.com/cloudposse/terraform-aws-cloudfront-cdn is not supporting terraform 0.12 . So I switch to use this module. And to host entire website. Use S3 static website as origin is much better to serve entire website. There is an old issue #14. But it didn't use the website URL as the origin. 

The main benefit is. When we have following files in S3:

```
- index.html
- photo/index.html
```

And when user navigate to `/photo` (without ending `/`). The S3 origin will return 404. But the S3 Static Website will return 302 redirect to `/photo/`.

Which is documented in last part of: https://docs.aws.amazon.com/AmazonS3/latest/dev/IndexDocumentSupport.html

> However, if you exclude the trailing slash from the preceding URL, Amazon S3 first looks for an object photos in the bucket. If the photos object is not found, then it searches for an index document, photos/index.html. If that document is found, Amazon S3 returns a 302 Found message and points to the photos/ key. For subsequent requests to photos/, Amazon S3 returns photos/index.html. If the index document is not found, Amazon S3 returns an error. 

And the redirection is also very helpful:

* https://docs.aws.amazon.com/AmazonS3/latest/user-guide/redirect-website-requests.html
* https://docs.aws.amazon.com/AmazonS3/latest/dev/how-to-page-redirect.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ Available targets:
 | forward_query_string | Forward query strings to the origin that is associated with this cache behavior | bool | `false` | no |
 | geo_restriction_locations | List of country codes for which  CloudFront either to distribute content (whitelist) or not distribute your content (blacklist) | list(string) | `<list>` | no |
 | geo_restriction_type | Method that use to restrict distribution of your content by country: `none`, `whitelist`, or `blacklist` | string | `none` | no |
-| index_document | Amazon S3 returns this index document when requests are made to the root domain or any of the subfolders | string | `` | no |
+| index_document | Amazon S3 returns this index document when requests are made to the root domain or any of the subfolders | string | `index.html` | no |
 | ipv6_enabled | Set to true to enable an AAAA DNS record to be set as well as the A record | bool | `true` | no |
 | lambda_function_association | A config block that triggers a lambda function with specific actions | object | `<list>` | no |
 | log_expiration_days | Number of days after which to expunge the objects | number | `90` | no |
@@ -204,6 +204,7 @@ Available targets:
 | viewer_protocol_policy | allow-all, redirect-to-https | string | `redirect-to-https` | no |
 | wait_for_deployment | When set to 'true' the resource will wait for the distribution status to change from InProgress to Deployed | bool | `true` | no |
 | web_acl_id | ID of the AWS WAF web ACL that is associated with the distribution | string | `` | no |
+| website_enabled | Set to true to use an S3 static website as origin | bool | `false` | no |
 
 ## Outputs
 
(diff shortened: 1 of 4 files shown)


aknysh:
@othree  thanks!
I think it was already implemented in https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/releases/tag/0.15.0 and is already in `master`

othree:
Hi, @aknysh 0.15 only includes configs related to S3 Static Website. The origin are still using a **S3** domain. Not a **S3 Static Website** domain. These features are only available under **S3 Static Website** domain

othree:
Hi, @aknysh 0.15 only includes configs related to S3 Static Website. The origin are still using a **S3** domain. Not a **S3 Static Website** domain. These features are only available under **S3 Static Website** domain

aknysh:
@othree  ok I see.
The module already has a local var for `website_enabled` https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/commit/2ec1ef6ceb2ba56e1ffa0ad8d6509b5dd88a2113#diff-7a370d8342e7203b805911c92454f0f4R2
You are introducing another one `var.website`.
Can you please take a look at that and consolidate what was done in 0.15.0 and in your PR?
Thanks

othree:
Do you prefer automatically detection enable or not? If yes I can use it instead.

aknysh:
I'd prefer an explicit var, like you did, but maybe with the name `var.website_enabled` (easier to understand what it does by just looking at the name)).
The automatic detection logic is already convoluted and not easy to read/understand.

othree:
Updated and use `website_enabled` instead. I am testing it again. Which take some time.

othree:
@aknysh Anything I can do to improve this or do you have any other idea?

bionicles:
this is a common use case for react devs and would be handy to have this flag available in master

bionicles:
be sure to test the DNS -- a good static site works with:

http://example.com
http://www.example.com
https://example.com
https://www.example.com

basically, it needs to work with custom domains on both HTTP and HTTPS with and without the WWW ... this can take some tinkering and may actually require to use the bucket domain and NOT the bucket website domain, and multiple aliases for www -> cloudfront and then alias for root -> www.

Definitely a headache for many

br4in3x:
@othree Didn't see your PR and created my own fork. Pls take a look, could be a little lighter: https://github.com/br4in3x/terraform-aws-cloudfront-s3-cdn/pull/1

othree:
@br4in3x I think you are right about the variables. I will update it. And does your fork works correctly? AFAIK the S3 have to use public policy(readable by everyone). Otherwise the website will not be viewable.

br4in3x:
@othree Yes, you're right. The bucket should be public, for sure. Thank you!

othree:
@maximmi I am very sorry I didn't push the correct commit last time. Now the documents are fixed. I will find a way to run test on my Mac to confirm it.(It throws error because of awk not compatiable)

PR #80, opened 2020-05-04
add output value to expose internal logs module in so that the logs

pingwin:
output can be used in parent terraform manifests

## what
* Describe high-level what changed as a result of these commits (i.e. in plain-english, what do these changes mean?)
* Use bullet points to be concise and to the point.

When working with cloudfront the output of the logs are an important feature. By exposing the logs module an implementer can now depend_on the s3 bucket and add (for example) aws_s3_bucket_notifications.

## why
* Provide the justifications for the changes (e.g. business case). 
* Describe why these changes were made (e.g. why do these commits fix the problem?)
* Use bullet points to be concise and to the point.

## references
* Link to any supporting github issues or helpful documentation to add some context (e.g. stackoverflow). 
* Use `closes #123`, if this PR closes a GitHub issue `#123`


Gowiem:
@aknysh @osterman What're thoughts on outputting a whole module like this? Do we want to support doing this or would you folks rather that we keep it explicit with only exporting out specific attributes?

pingwin:
If that's the approach wanted, I can change the code to export ARNs and IDs instead.

nitrocode:
Oh wow I didn't realize you can do this.

I like the idea of outputting the entire module since we don't have to expose its individual outputs.

PR #81, opened 2020-05-08
Output the ARN of the Origin S3 Bucket

tampakrap:
## what
* Output the ARN of the Origin S3 Bucket

## why
* This is useful in cases where a policy is created a CI/CD user or role,
  which needs access to the Origin S3 Bucket in order to add content.

## references
* None


Gowiem:
@tampakrap Looks like your code failed our terratests and in a legitimate way: 

```
Because aws_s3_bucket.origin has "count" set, its attributes must be accessed on specific instances.
```

[You'll need to do access the arn like so](https://github.com/cloudposse/terraform-aws-vpc/blob/master/outputs.tf#L6-L8): 

```
value       = join("", aws_s3_bucket.origin.*.arn)
```

Can you make that update? Also, your code failed `terraform validate`. Mind running `terraform fmt` and `terraform validate` as well before pushing again?

Make those changes and we'll get that merged -- Thanks man!

tampakrap:
@nitrocode thank you for doing that for me!
@Gowiem sorry for the delay and thanks for the review

PR #84, opened 2020-06-05
Dynamically generate cors rules, fixes #22

### LINKED ISSUE #22: Bug in configuration of allowed origins for S3 bucket
According to [S3 CORS configuration Amazon documentation](https://docs.aws.amazon.com/AmazonS3/latest/dev/cors.html) there is a bug in configuration of allowed origins.

Example from documentation shows that each CORS rule need to consist only one AllowedOrigin parameter:
```
<CORSConfiguration>
 <CORSRule>
   <AllowedOrigin>http://www.example1.com</AllowedOrigin>

   <AllowedMethod>PUT</AllowedMethod>
   <AllowedMethod>POST</AllowedMethod>
   <AllowedMethod>DELETE</AllowedMethod>

   <AllowedHeader>*</AllowedHeader>
 </CORSRule>
 <CORSRule>
   <AllowedOrigin>http://www.example2.com</AllowedOrigin>

   <AllowedMethod>PUT</AllowedMethod>
   <AllowedMethod>POST</AllowedMethod>
   <AllowedMethod>DELETE</AllowedMethod>

   <AllowedHeader>*</AllowedHeader>
 </CORSRule>
 <CORSRule>
   <AllowedOrigin>*</AllowedOrigin>
   <AllowedMethod>GET</AllowedMethod>
 </CORSRule>
</CORSConfiguration>
```

In case if assign `example.com` value  to the `cors_allowed_origins` variable, result will be next:
```
<CORSConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
<CORSRule>
    <AllowedOrigin>example.com</AllowedOrigin>
    <AllowedOrigin>some.otherdomain.com</AllowedOrigin>
    <AllowedMethod>GET</AllowedMethod>
    <MaxAgeSeconds>3600</MaxAgeSeconds>
    <ExposeHeader>ETag</ExposeHeader>
    <AllowedHeader>*</AllowedHeader>
</CORSRule>
</CORSConfiguration>
```

This is the cause of "floating" bug, when sometimes calls works fine, but other times returns a CORS error.

Makeshift:
## what
* We now dynamically generate the cors rules block so we correctly create a block for each alias/allowed origin.

## why
* Cors origins don't work without these changes
* I believe this change makes them work as intended in the original module

## references
* Close #22 
* I hit enter too early, sorry for the edits :)


osterman:
Unfortunately tests are failing b/c someone's tests have created the bucket and not deleted it. We'll need to fix the tests in this module before merging by generating the bucket name. 

If you are able to incorporate these changes, that would be much appreciated. See example here: https://github.com/cloudposse/terraform-aws-alb/pull/42/files

Gowiem:
@Makeshift Thanks for the contribution! Released as [0.24.0](https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/releases/tag/0.24.0).

PR #91, opened 2020-07-23
Adding versioning to S3 origin bucket

hawkesnc:
## what
* This change allows you to turn on versioning for objects in the bucket

## why
* For building static websites, we require the use of s3+cdn that this module provides. However, in some instances, we would like to ability to roll back to a previous version. 
* Since this is meant for a "dev" like environment, we set the cdn ttl to very low to pull objects back from the s3 bucket.
* The reason we use the CDN is to support TLS since s3 does not.

## references
* N/A

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -124,10 +124,10 @@ resource "aws_s3_bucket" "origin" {
   }
 
   dynamic "versioning" {
-    for_each = var.versioning_enabled ? ["true"] : []
+    for_each = [var.versioning_enabled]
 
     content {
-      enabled = var.versioning_enabled
+      enabled = versioning.value
     }
   }
 
--- a/main.tf
+++ b/main.tf
@@ -123,6 +123,14 @@ resource "aws_s3_bucket" "origin" {
     }
   }
 
+  dynamic "versioning" {
+    for_each = var.versioning_enabled ? ["true"] : []
+
+    content {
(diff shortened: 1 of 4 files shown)


main.tf
```diff
@@ -123,6 +123,14 @@ resource "aws_s3_bucket" "origin" {
     }
   }
 
+  dynamic "versioning" {
+    for_each = var.versioning_enabled ? ["true"] : []
```
nitrocode, line 127: This can be simplified to this
```suggestion
    for_each = [var.versioning_enabled]
```

main.tf
```diff
@@ -123,6 +123,14 @@ resource "aws_s3_bucket" "origin" {
...
+
+    content {
+      enabled = var.versioning_enabled
```
nitrocode, line 130:
```suggestion
      enabled = versioning.value
```

hawkesnc:
Added those changes in. Thanks @nitrocode for the review! 👍

hawkesnc:
Is this good to merge?

hawkesnc:
Sweet! Thanks 👍

Gowiem:
@hawkesnc Thanks for the contribution! Released as [0.34.0](https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/releases/tag/0.34.0).

PR #99, opened 2020-08-23
Remove the v prefix

nitrocode:
## what
Remove the v prefix to keep releases consistent

## why
* Consistency

## references
* @Makeshift pointed this out earlier today

### DIFF
--- a/.github/auto-release.yml
+++ b/.github/auto-release.yml
@@ -1,4 +1,4 @@
-name-template: 'v$RESOLVED_VERSION'
+name-template: '$RESOLVED_VERSION'
 tag-template: '$RESOLVED_VERSION'
 version-template: '$MAJOR.$MINOR.$PATCH'
 version-resolver:


Makeshift:
> The tag name is correct - this is just the title.

This is why I labelled it as a minor inconsistency in chat - It confused me for a couple of seconds because I accidentally wrote the release name and not the tag name when doing a `?ref` (because I'm used to them being the same). But it's a cosmetic issue really, it doesn't make a difference unless people don't know how `?ref` works.

I'm assuming the correct place to PR this would be [here](https://github.com/cloudposse/build-harness/blob/f28afcf5ca7bf4cc27de86099d8bd701899e7736/templates/.github/auto-release.yml#L1) and [here](https://github.com/cloudposse/build-harness/blob/f28afcf5ca7bf4cc27de86099d8bd701899e7736/.github/auto-release.yml#L1), though I'm not sure it even merits the effort ;)

nitrocode:
Oh woops. Good catch folks.

PR #100, opened 2020-09-02
use minimum versions for provider pinning

woz5999:
## what
set minimum versions for providers without pinning to a specific major version

## why
the current method makes it very difficult to maintain or upgrade consistent provider versions within a project

## references
https://www.terraform.io/docs/configuration/provider-requirements.html#best-practices-for-provider-versions

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -364,7 +364,7 @@ resource "aws_cloudfront_distribution" "default" {
 }
 
 module "dns" {
-  source           = "git::https://github.com/cloudposse/terraform-aws-route53-alias.git?ref=tags/0.8.0"
+  source           = "git::https://github.com/cloudposse/terraform-aws-route53-alias.git?ref=tags/0.9.0"
   enabled          = var.enabled && (var.parent_zone_id != "" || var.parent_zone_name != "") ? true : false
   aliases          = var.aliases
   parent_zone_id   = var.parent_zone_id
--- a/main.tf
+++ b/main.tf
@@ -156,7 +156,7 @@ resource "aws_s3_bucket" "origin" {
 }
 
 module "logs" {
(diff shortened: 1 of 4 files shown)


main.tf
```diff
@@ -364,7 +364,7 @@ resource "aws_cloudfront_distribution" "default" {
 }
 
 module "dns" {
-  source           = "git::https://github.com/cloudposse/terraform-aws-route53-alias.git?ref=tags/0.8.0"
+  source           = "git::https://github.com/cloudposse/terraform-aws-route53-alias.git?ref=tags/0.9.0"
```
woz5999, line 367: per https://github.com/cloudposse/terraform-aws-route53-alias/pull/25

woz5999:
@aknysh

main.tf
Nuru, line 367:
```suggestion
  source           = "git::https://github.com/cloudposse/terraform-aws-route53-alias.git?ref=tags/0.8.2"
```

Nuru:
@osterman AWS provider 3.0 has  a breaking change that breaks this module. `aws_s3_bucket` [2.x](https://registry.terraform.io/providers/hashicorp/aws/2.70.0/docs/resources/s3_bucket#region) allows setting the region on the resource, while 3.x forbids it. How do you want to handle this?

woz5999:
@Nuru if it's helpful, i've got an aws 3.0 PR open for the s3 bucket module

osterman:
Ya all of our bucket modules have problem like the tf state backend module. I say bite the bullet and upgrade. However the fix is just to remove the region and the region passed with the provider should work for both versions, no?

woz5999:
yeah that's correct. and anyone with a corner case who might have been specifying a bucket region different from the provider would get an error on upgrade rather than tf accidentally trying to make some kind of destruction change

Nuru (approved):

PR #102, opened 2020-09-10
Toggling Route53 Entries

hawkesnc:
## what
This PR allows you to toggle the creation of the Route53 entries based on the alias list

## why
In certain situations, you may want to add the alternative CNAMEs to the CDN without creating the corresponding route53 entries. 

For example, if you wanted to create an alias but use a different record type (ex. weighted). The current behaviour would fail to create route53 records if you already have an existing route53 record.

I would prefer to be able to turn the route53 entries off and use another module to handle the route53 entries.


hawkesnc:
hey guys! Any update on this one? 😄  Thank you 👍

hawkesnc:
Closing in favour of new `0.35.0` features

PR #113, opened 2020-11-17
add option for aws_s3_bucket_public_access_block for origin

juliusdanielherreraglomm:
## what
* This change allows you to add a `aws_s3_bucket_public_access_block`
resource for the origin s3 bucket.

## why
* On "Amazon S3 Preventative Security Best Practices" it is recommended
to use Amazon S3 block public access.

## references
* https://docs.aws.amazon.com/AmazonS3/latest/dev/security-best-practices.html

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -146,6 +146,15 @@ resource "aws_s3_bucket" "origin" {
   }
 }
 
+resource "aws_s3_bucket_public_access_block" "origin" {
+  count                   = ! local.using_existing_origin && var.block_origin_public_access_enabled ? 1 : 0
+  bucket                  = local.bucket
+  block_public_acls       = true
+  block_public_policy     = true
+  ignore_public_acls      = true
+  restrict_public_buckets = true
+}
+
 module "logs" {
   source                   = "git::https://github.com/cloudposse/terraform-aws-s3-log-storage.git?ref=tags/0.15.0"
   enabled                  = var.logging_enabled
--- a/variables.tf
+++ b/variables.tf
@@ -429,3 +429,9 @@ variable "origin_ssl_protocols" {
   default     = ["TLSv1", "TLSv1.1", "TLSv1.2"]
   description = "The SSL/TLS protocols that you want CloudFront to use when communicating with your origin over HTTPS."
 }
+
+variable "block_origin_public_access_enabled" {
+  type        = bool
+  default     = false
+  description = "When set to 'true' the s3 origin bucket will have public access block enabled"
+}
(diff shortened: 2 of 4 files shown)


Gowiem:
@juliusdanielherreraglomm can you rebase off master to fix these conflicts and then we'll get this merged? Thanks!

variables.tf
```diff
@@ -417,3 +417,9 @@ variable "cloudfront_origin_access_identity_path" {
   default     = ""
   description = "Existing cloudfront origin access identity path used in the cloudfront distribution's s3_origin_config content"
 }
+
+variable "origin_public_access_block" {
```
Gowiem, line 421: @juliusdanielherreraglomm can you update this variable name to `block_origin_public_access_enabled` or something similar? We use the `_enabled` convention to denote `bool` var where we can.

Gowiem (changes requested):

variables.tf
juliusdanielherreraglomm: Good idea. Done.

juliusdanielherreraglomm: Hi @Gowiem

Happy new year!

I tried include your suggestion and rebased it. I hope that this is now ready to be merged.
Let me know if I can do more.

Cheers

Julius

main.tf
```diff
@@ -146,6 +146,15 @@ resource "aws_s3_bucket" "origin" {
   }
 }
 
+resource "aws_s3_bucket_public_access_block" "origin" {
+  count                   = ! local.using_existing_origin && var.block_origin_public_access_enabled ? 1 : 0
+  bucket                  = module.origin_label.id
```
Gowiem, line 151: @juliusdanielherreraglomm apologies I didn't catch this before, but not using the `aws_s3_bucket.origin.id` (the bucket name) here might cause this resource to try and be created before the `aws_s3_bucket.origin` resource which would result in an error. Can you update this to use the reference to the bucket name instead of the `origin_label.id`?

juliusdanielherreraglomm: @Gowiem Good catch. This similar to the `aws_s3_bucket_policy.default` resource which also uses the bucket name and needs to ensure that the bucket is created before. In this case it is ensured by using `local.bucket` which has a dependency to the bucket.

For consistency reasons I suggest to use the same way. I modified the PR. Could you take a look?

Gowiem (approved):

main.tf
Gowiem: Looks good @juliusdanielherreraglomm -- Thanks for the update.

Issue #115, opened 2020-12-09
Add output for Cloudfront Distribution Origin Access Identity


bkero:
Have a question? Please checkout our [Slack Community](https://slack.cloudposse.com) or visit our [Slack Archive](https://archive.sweetops.com/). 

[![Slack Community](https://slack.cloudposse.com/badge.svg)](https://slack.cloudposse.com)

## Describe the Feature

Provide access to the IAM ARN of the Origin Access Identity of the Cloudfront Distribution.

## Expected Behavior

An output named `cf_identity_iam_arn` or similar should provide access to the IAM ARN of this resource. If the resource doesn't exist it should fail to "" since conditional existence of outputs in Terraform is tricky.

## Use Case

We want to use an existing S3 bucket policy. To do that, but also provide access to the Cloudfront distribution created by this module, it is necessary to have access to that distribution's Origin Access Identity. 

## Describe Ideal Solution

PR incoming

## Alternatives Considered

Rip it out of the state file using a self-referential remote state data reference.

PR #116, opened 2020-12-09
Add cloudfront origin identity iam arn output

bkero:
This adds an output to access the Cloudfront distribution's Origin
Access identity, if it is enabled, else a blank string.

We need this to provide access to the Cloudfront distribution from an externally-written bucket policy.

Closes: #115

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,8 +64,15 @@ We literally have [*hundreds of terraform modules*][terraform_modules] that are 
 ## Usage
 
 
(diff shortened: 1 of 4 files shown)


jamengual:
@bkero check the tests please, this is not passing the tests

bkero:
@jamengual I believe I've fixed this in revised commit. Can you run the tests again?

Gowiem (approved):
LGTM

PR #136, opened 2021-03-08
Serialise origin bucket modifications

alexjurkiewicz:
You can't modify an S3 bucket's policy & public access block at the same
time, AWS API will complain:

OperationAborted: A conflicting conditional operation is currently in progress against this resource

This error can appear during both deploy and destroy for the module.

Serialise operations to the origin bucket so we don't run into this
error. The suggested fix is from

https://github.com/hashicorp/terraform-provider-aws/issues/7628

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -94,7 +94,7 @@ data "template_file" "default" {
 
 resource "aws_s3_bucket_policy" "default" {
   count  = ! local.using_existing_origin || var.override_origin_bucket_policy ? 1 : 0
-  bucket = aws_s3_bucket.origin.bucket
+  bucket = join("", aws_s3_bucket.origin.*.bucket)
   policy = data.template_file.default.rendered
 }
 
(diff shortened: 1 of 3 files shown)


main.tf
```diff
@@ -94,7 +94,7 @@ data "template_file" "default" {
 
 resource "aws_s3_bucket_policy" "default" {
   count  = ! local.using_existing_origin || var.override_origin_bucket_policy ? 1 : 0
-  bucket = local.bucket
+  bucket = aws_s3_bucket.origin.bucket
```
joe-niland, line 97:
```suggestion
  bucket = join("", aws_s3_bucket.origin.*.bucket)
```

joe-niland (changes requested):
@alexjurkiewicz just a small typo to fix

alexjurkiewicz:
Thank you! Fixed

joe-niland (approved):

joe-niland:
Thank you @alexjurkiewicz !

PR #140, opened 2021-03-17
Add support for a custom cache policy

dmattia:
## what

- AWS has a relatively new CloudFront feature named Cache Policies: https://aws.amazon.com/blogs/networking-and-content-delivery/amazon-cloudfront-announces-cache-and-origin-request-policies/
- This adds support for the default cache to set an explicit cache policy
- This also sets restrictions on not allowing some other params to be set if the cache policy is specified, as is consistent with the AWS Console's wizard when creating a distribution.

This module could also create a cache policy itself if desired, I would gladly make that change.

Note: I support a module for creating cache policies: https://registry.terraform.io/modules/transcend-io/cache-policy/aws/latest

## why

- Cache Policies are the easiest (only?) way to add brotli compression in CloudFront


main.tf
```diff
@@ -289,16 +289,21 @@ resource "aws_cloudfront_distribution" "default" {
...
+    cache_policy_id  = var.cache_policy_id
     target_origin_id = module.this.id
     compress         = var.compress
     trusted_signers  = var.trusted_signers
 
-    forwarded_values {
-      query_string = var.forward_query_string
-      headers      = var.forward_header_values
+    dynamic "forwarded_values" {
+      # If a cache policy is specified, we cannot includ a `forwarded_values` block at all in the API request
```
nitrocode, line 298:
```suggestion
      # If a cache policy is specified, we cannot include a `forwarded_values` block at all in the API request
```

justnom:
@dmattia 

Should we also add the `cache_policy_id` into the `ordered_cache` map?

dmattia:
We definitely could do that as well. Would you like me to make that change?

justnom:
> We definitely could do that as well. Would you like me to make that change?

Sure, or I can do it in a separate PR after this is merged. I can't seem to suggest changes to the PR from outside the scope shown in "Files changed".

dmattia:
@justnom Awesome, let's do the ordered caches in a separate PR then 👍

Gowiem:
@dmattia apologies for giving you conflicts yet again. This looks like a straight forward change... want to update once last time and then ping me and we'll get this merged for you?

nitrocode (approved):

Issue #90, opened 2020-07-20
S3 Bucket not found


JohnShortland:
## Describe the Bug
Ran using the example from repo and get the following error:

```
Error: Failed getting S3 bucket: NotFound: Not Found
        status code: 404, request id: EAC7428BD0E1699D, host id: rQ8gPW9bBhoBc6IOtIIslXHcioIr3VFwti9wxiPCZ6FqkUn/n2+qvqaVefuhcUzDteJdOavMLSU= Bucket: "BUCKET_NAME.DOMAIN.COM"

  on .terraform/modules/cloudfront_s3_cdn/main.tf line 176, in data "aws_s3_bucket" "selected":
 176: data "aws_s3_bucket" "selected" {
```

This is being ran on an organisation account in eu-west-1

## Expected Behavior
Bucket to be created and cloudfront distro created

## Screenshots

```hcl
module "cloudfront_s3_cdn" {
  source                   = "git::https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn.git?ref=tags/0.26.0"
  namespace                = "${terraform.workspace}"
  stage                    = var.stage
  name                     = "BUCKET_NAME"
  use_regional_s3_endpoint = true
  origin_force_destroy     = true
  cors_allowed_headers     = ["*"]
  cors_allowed_methods     = ["GET", "HEAD", "PUT"]
  cors_allowed_origins     = ["*.${var.domain}"]
  cors_expose_headers      = ["ETag"]
  acm_certificate_arn      = aws_acm_certificate.BUCKET_NAME.arn
  aliases                  = ["BUCKET_NAME.${var.domain}"]
  origin_bucket            = "BUCKET_NAME.${var.domain}"
}
```

## Environment (please complete the following information):
 - OS: WSL
 - Version: 2.0

shushry:
I ran into this also - I think there is some confusion about the parameter "origin_bucket".   If you were like me, you were assuming you could define the name of a NEW bucket to contain your origin, using that parameter.   However, the module wants to name a new bucket FOR you, dynamically, based on the values of namespace, stage, etc.   If you specify "origin_bucket" it's looking for an EXISTING bucket that you might already have out there.    

Docs should be updated to clarify this.

nitrocode:
Yes, the `origin_bucket` is for an existing s3 bucket. I put in a PR to update the description of this variable.

PR #142, opened 2021-03-27
Provide better description for var.origin_bucket

nitrocode:
## what
* Provide better description for var.origin_bucket
* Provide an example of reusing an s3 bucket in the README

## why
* There's some confusion whether to define this value as the bucket to be created or to reuse an existing bucket

## references
* Closes https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/issues/90


variables.tf
```diff
@@ -49,7 +49,7 @@ variable "override_origin_bucket_policy" {
 variable "origin_bucket" {
   type        = string
   default     = ""
-  description = "Origin S3 bucket name"
+  description = "Use an existing origin S3 bucket name"
```
osterman, line 52: Name of an existing S3 bucket to use as the origin.

nitrocode: Good comment. I updated the description with that and more.

osterman (dismissed):

jamengual (approved):

Issue #101, opened 2020-09-10
BucketRegionError: incorrect region


thomas-beznik:
## Describe the Bug
I am getting the following error when using this module:
```
Error: Failed getting S3 bucket: BucketRegionError: incorrect region, the bucket is not in 'eu-central-1' region at endpoint ''
        status code: 301, request id: , host id:  Bucket: "aws-cli"
```
this error doesn't occur anymore if I delete my `.terraform` folder and the `terraform.tfstate` file.

## Steps to Reproduce
This is the configuration that I'm using:

```hcl
provider "aws" {
  region = var.region
}

/*
* CDN 
*/
module "cloudfront_s3_cdn" {
  acm_certificate_arn           = var.certificate_arn
  aliases                       = ["${var.name}.${var.environment}.toothflow.com"]
  source                        = "cloudposse/cloudfront-s3-cdn/aws"
  version                       = "0.34.0"
  stage                         = var.environment
  name                          = "${var.name}"
  override_origin_bucket_policy = true
  default_root_object           = "index.html"
  compress                      = true
  price_class                   = "PriceClass_100"
  use_regional_s3_endpoint      = true 
  origin_force_destroy          = true 
  cors_allowed_headers          = ["*"]
  cors_allowed_methods          = ["GET", "HEAD"]
  cors_allowed_origins          = ["${var.name}.${var.environment}.toothflow.com"]
  cors_expose_headers           = ["ETag"]
  viewer_protocol_policy        = "redirect-to-https"
}
```

Additional
```hcl
/*
* S3 Bucket 
*/
resource "aws_s3_bucket_object" "bucket-s3" {
  bucket       = module.cloudfront_s3_cdn.s3_bucket
  key          = "${var.name}-${var.environment}"
  content_type = "text/html"
  acl          = "public-read"
}

/*
* Route53 Record 
*/
resource "aws_route53_record" "route53-record" {
  zone_id         = var.S3_zone_id
  name            = "${var.domain}.${var.environment}.toothflow.com"
  type            = "CNAME"
  ttl             = "300"
  records         = [module.cloudfront_s3_cdn.cf_domain_name]
  allow_overwrite = true
}

/*
* IAM Programmatic Access to S3 Bucket 
*/

// IAM User
resource "aws_iam_user" "S3_access_user" {
  name = "S3-${var.name}-${var.environment}"
  path = "/"
  force_destroy = true
}

// IAM Policy
[...]

PR #143, opened 2021-03-27
Remove static bucket hack

nitrocode:
This is for comments to ensure this PR doesn't break backwards compatibility.

## what
* This simplifies calculating the correct bucket regional domain name using the s3 bucket's `bucket_regional_domain_name` instead of the original hack from PR https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/17
* I saw this output used on the official hashicorp docs for the [aws_cloudfront_distribution](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution) resource

## why
* This simplifies the regional domain name
* Prevents having to maintain a list of buckets with dashes
* Prevents issues related to the data source of the `aws-cli` s3 bucket
* Should maintain destruction of the resources as it does before

## references
* More digging down the rabbit hole in this ticket https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/issues/101


aknysh (dismissed):
@nitrocode 
Thanks, LGTM.
I think `var.use_regional_s3_endpoint` is not in use anymore and needs to be removed from 

https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/blob/master/variables.tf#L31

Also, please run 
```
make init
make github/init
make readme
```

and commit the changes.

Thanks

aknysh (approved):
thanks @nitrocode

main.tf
```diff
@@ -190,13 +181,7 @@ locals {
       concat([var.origin_bucket], concat([""], aws_s3_bucket.origin.*.id))
     )
   )
-
-  bucket_domain_name = (var.use_regional_s3_endpoint || var.website_enabled) ? format(
-    var.website_enabled ? "%s.s3-website%s%s.amazonaws.com" : "%s.s3%s%s.amazonaws.com",
```
syphernl, line 195: @nitrocode This change seems to have broken the "s3-website" functionality. It now wants to change the origin `domain_name` from 
`XXXXX.s3-website.eu-west-1.amazonaws.com` to `XXXXX.s3.eu-west-1.amazonaws.com` which results in an `AccessDenied` when loading the domain.

nitrocode: Ah what we want is bucket regional domain name output of the bucket. I'll put in another pr

nitrocode: I created this ticket https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/issues/148 after reviewing your comment. Good catch.

alexjurkiewicz:
The removal of `use_regional_s3_endpoint` parameter was also a breaking change, and not communicated in release notes :(

PR #147, opened 2021-03-30
Correct enabled behavior

syphernl:
## what
* Replaced the `enabled` flag with `distribution_enabled`
* Corrected behavior for actual `enabled` flag to prevent from creating any resources

## why
* The `enabled` input was being used to manage the distribution state, which does not match the description of "Set to false to prevent the module from creating any resources" since setting `enabled` to `false` still resulted in the CloudFront distribution and S3 bucket to be created.
* The behavior is now similar to that of [terraform-aws-cloudfront-cdn](https://github.com/cloudposse/terraform-aws-cloudfront-cdn)


jamengual:
@syphernl please check the test, thanks

main.tf
```diff
@@ -188,8 +192,10 @@ locals {
 }
 
 resource "aws_cloudfront_distribution" "default" {
+  count = module.this.enabled ? 1 : 0
+
   #bridgecrew:skip=BC_AWS_LOGGING_20:Skipping `CloudFront Access Logging` check until bridgecrew will support dynamic blocks (https://github.com/bridgecrewio/checkov/issues/776).
-  enabled             = module.this.enabled
+  enabled             = var.distribution_enabled
```
jamengual, line 198: I think this should be `count  = module.this.enabled && var.distribution_enabled` otherwise you could have a bucket without a cloudfront distribution and that is not the purpose of this module.

syphernl: The `enabled` you referred to is whether the CF Distribution needs to be enabled or not. The count is for that resource is `count = module.this.enabled ? 1 : 0` , which prevents the CF distribution from being created when the cf-s3-cdn module is set to `enabled = false`.

If we change the `count` for this module to the one you mentioned, we cannot toggle the status of the CF Distribution itself (on AWS side).

jamengual: This module primarily created a CF distribution and a public s3 bucket ( plus some other stuff like logs, permissions etc) so if we allow for the CF distribution to be disabled, then we are creating a public bucket basically and that is not what the module is meant for, I think

jamengual: I see ok, sorry it is very rare to have an `enabled` flag on a resource.

jamengual (approved):

main.tf
```diff
@@ -73,7 +77,7 @@ data "aws_iam_policy_document" "origin_website" {
 }
 
 resource "aws_s3_bucket_policy" "default" {
-  count  = ! local.using_existing_origin || var.override_origin_bucket_policy ? 1 : 0
+  count  = (module.this.enabled && (! local.using_existing_origin || var.override_origin_bucket_policy)) ? 1 : 0
```
jamengual, line 80: @syphernl I think this is causing issues I think it should be like : 
```
count  = (module.this.enabled && ( local.using_existing_origin || var.override_origin_bucket_policy)) ? 1 : 0
```

Issue #153, opened 2021-04-16
Variables in additional_bucket_policy are not substituted anymore


rafaljanicki:
## Describe the Bug
The documentation states that the additional_bucket_policy variable supports variables inside such as `${bucket_name}`. However, that doesn't seem to be the case anymore and the variable is passed to Terraform without any replace done. That's probably done because of the lack of template_file

## Expected Behavior
`${bucket_name}` should be replaced to match the bucket's name

## Steps to Reproduce
```
data "aws_iam_policy_document" "document" {
  dynamic "statement" {
    for_each = var.accounts

    content {
      sid = "JustASide${each.key}"

      actions = ["s3:GetObject"]
      resources = [
        "arn:aws:s3:::$${bucket_name}",
        "arn:aws:s3:::$${bucket_name}/*"
      ]

      principals {
        type = "AWS"
        identifiers = ["arn:aws:iam::${statement.value}:root"]
      }
    }
  }
}

module "cloudfront-s3-apigateway-cdn" {
  source = "cloudposse/cloudfront-s3-cdn/aws"
  version = "~> 0.59"

  compress = true
  dns_alias_enabled = true
  price_class = "PriceClass_All"

  acm_certificate_arn = var.static_acm_certificate_arn
  aliases = ["static.${var.static_domain_name}"]
  block_origin_public_access_enabled = true
  environment = var.environment
  name = "static"
  origin_force_destroy = true
  origin_ssl_protocols = ["TLSv1.2"]
  parent_zone_id = var.static_domain_name_zone_id
  website_enabled = true

  additional_bucket_policy = data.aws_iam_policy_document.document.json

  tags = local.common_tags
}
```

## Environment:

Anything that will help us triage the bug will help. Here are some ideas:
 - OS: OSX
 - Module version: 0.59

## Additional context

The behaviour was removed with this commit: https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/commit/c7e6f3e6d1513157ca4c4e05379e39bd4dc20409

PR #158, opened 2021-05-11
`var.additional_bucket_policy` should overwrite vars again

nitrocode:
## what
* Use templatefile with previous variables for the additional bucket policy

## why
* Allows `var.additional_bucket_policy` to take advantage of the original variables that were replaced prior to PR https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/141

## references
* Closes https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/issues/153


Nuru (changes requested):
Let's not use a template file if we don't have to.

@nitrocode Please pause work on this PR. I expect to fix #153 in an upcoming PR soon

PR #105, opened 2020-10-15
Make S3 website private; add S3 origin

jmcgeheeiv:
## What

In this PR I added these enhancements:

- Make the S3 website accessible only via CloudFront. You may prefer that I add a variable to make this optional. Let me know.
- Add S3 origins, which are different from the currently-supported custom origins
- Change `http_port` and `https_port` to numbers merely to dissuade observers from attempting to use `null`. `null` causes a very, very hard-to-diagnose error that does not reveal the line number.

I will further note that Terraform insists that the complete data structure appearing in `variables.tf` be specified for `custom_origin` and `s3_origin` variables. Thus `lookup()` with default values is immaterial.

## Why

The reasons are included in the list of enhancements above. I hope it is more readable that way.

PR #159, opened 2021-05-12
fix: only create referer password when needed

syphernl:
## what
* Only generate the random `referer` when website functionality is being used.

## why
* To prevent unnecessary resources that are are not always required.

## references
* #105

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -29,7 +29,7 @@ resource "aws_cloudfront_origin_access_identity" "default" {
 }
 
 resource "random_password" "referer" {
-  count = (module.this.enabled && var.website_enabled) ? 1 : 0
+  count = module.this.enabled && var.website_enabled ? 1 : 0
 
   length  = 32
   special = false
--- a/main.tf
+++ b/main.tf
@@ -29,6 +29,8 @@ resource "aws_cloudfront_origin_access_identity" "default" {
 }
 
 resource "random_password" "referer" {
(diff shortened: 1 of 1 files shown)


Nuru (changes requested):
This partly conflicts with #160. I will add this functionality to #160 instead of trying to merge this. I hope you don't mind, @syphernl

main.tf
```diff
@@ -29,6 +29,8 @@ resource "aws_cloudfront_origin_access_identity" "default" {
 }
 
 resource "random_password" "referer" {
+  count = (module.this.enabled && var.website_enabled) ? 1 : 0
```
nitrocode, line 32:
```suggestion
  count = module.this.enabled && var.website_enabled ? 1 : 0
```

main.tf
```diff
@@ -81,7 +83,7 @@ data "aws_iam_policy_document" "origin_website" {
     condition {
       test     = "StringEquals"
       variable = "aws:referer"
-      values   = [random_password.referer.result]
+      values   = [random_password.referer[0].result]
```
nitrocode, line 86: This will fail if `module.this.enabled = true` and `var.website_enabled = false` because `random_password.referer[0]` will not exist

nitrocode:
It's probably best to wait until this PR https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/160/files goes through

Nuru:
Superseded by #160

PR #162, opened 2021-05-13
Add support for realtime log config

linhkikuchi:
## what
* To add support for realtime log config for default cache


nitrocode (approved):
LGTM

PR #164, opened 2021-05-14
Fix: Deployment Principal ARNs

korenyoni:
## what
* Add test for `deployment_principal_arns`
* Fix order of nested functions when rendering bucket policy for `deployment_principal_arns` — otherwise an error occurs when comparing types.
* ~Fix snippet: all prefixes must begin in forward slash or otherwise the bucket policy JSON will be invalid.~
* Fix snippet passing incorrect variable name.
* Programatically fix missing leading forward slash for prefixes in Bucket Policy

### (misc)
* ~Fix typo in tfvars fixtures in example.~
* Add enablement toggles to all resources in example
* Fix minimum_protocol_version being passed when it won't be honored by AWS i.e. when s3_access_log_bucket_name is true as a result of acm_certificate_arn being unset.
* Fix default for s3_access_log_bucket_name — it is supposed to be empty and there is a check for its length in locals{} that assumes it is empty by default, but it is not empty by default.

## why
* `deployment_principal_arns` was not working and `examples/complete` was not testing for it

## references
* #160


examples/complete/fixtures.us-east-2.tfvars
```diff
@@ -6,4 +6,4 @@ stage = "test"
 
 name = "cloudfront-s3-cdn"
 
-parent_zone_name = "testing.cloudposse.co"
\ No newline at end of file
+parent_zone_name = "testing.cloudposse.com"
```
Nuru, line 9: Testing domain is `.co` _[sic]_
```suggestion
parent_zone_name = "testing.cloudposse.co"

```

examples/complete/main.tf
```diff
@@ -22,7 +30,27 @@ data "aws_iam_policy_document" "document" {
...
+
+  name = each.value
+
+  assume_role_policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Action    = "sts:AssumeRole"
+        Effect    = "Allow"
+        Sid       = ""
```
Nuru, line 48:
```suggestion
        Sid       = "EnableTestRole${each.key}"
```

jamengual (approved):

main.tf
```diff
@@ -382,7 +382,7 @@ resource "aws_cloudfront_distribution" "default" {
   viewer_certificate {
     acm_certificate_arn            = var.acm_certificate_arn
     ssl_support_method             = var.acm_certificate_arn == "" ? "" : "sni-only"
-    minimum_protocol_version       = var.minimum_protocol_version
+    minimum_protocol_version       = var.acm_certificate_arn == "" ? "TLSv1" : var.minimum_protocol_version # If cloudfront_default_certificate=true, only TLSv1 can be specified (see: https://docs.aws.amazon.com/cli/latest/reference/cloudfront/create-distribution.html).
```
Nuru, line 385: This should be documented in the `minimum_protocol_version` variable description, not fixed here. Fixing it here leaves users having set `minimum_protocol_version ="TLSv1.2"` thinking they have it set but secretly not.

Nuru (changes requested):

korenyoni:
Tests are currently stuck in queued due to a GitHub Actions incident

korenyoni:
Will need to remove validation due to Terraform version used in tests being too old for functions used in validation block.

Nuru (dismissed):

korenyoni:
@Nuru thank you — yes it does seem better to programatically handle the leading slash and not have the user deal with it. I just fixed a couple of typos in `deployment.tf`

main.tf
```diff
@@ -167,9 +170,9 @@ data "aws_iam_policy_document" "deployment" {
   statement {
     actions = var.deployment_actions
 
-    resources = flatten(distinct([
-      local.origin_bucket.arn,
-      formatlist("${local.origin_bucket.arn}%s*", each.value),
+    resources = distinct(flatten([
+      [local.origin_bucket.arn],
+      formatlist("${local.origin_bucket.arn}/%s*", each.value),
```
korenyoni, line 175: This is a better way of ensuring the bucket policy is valid, as we're handling the slash programatically now instead of deferring it to the user.

aknysh (approved):

mcalhoun (commented):
LGTM!

PR #166, opened 2021-05-17
Use bucket.website_endpoint instead of bucket.website_domain

Zyava:
## what
There is currently a bug introduced in version v0.64.0 which prevents CloudFront from connecting to the bucket. `bucket.website_domain = "s3-website-<region>.amazonaws.com"` (see AWS provider source code https://github.com/hashicorp/terraform-provider-aws/blob/ac06ced75cba0daf09fef2938752ad13cc6fff6e/aws/resource_aws_s3_bucket.go#L1747). In https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/blob/master/main.tf#L319 `local.bucket_domain_name` is used in CloudFront origin, but since it is  "s3-website.eu-central-1.amazonaws.com" (in our case) - CloudFront gets a redirect to https://aws.amazon.com/s3/ and nothing works...

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -47,7 +47,7 @@ locals {
   ) : ""
 
   bucket             = local.origin_bucket.bucket
-  bucket_domain_name = var.website_enabled ? local.origin_bucket.website_domain : local.origin_bucket.bucket_regional_domain_name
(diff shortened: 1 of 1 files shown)


korenyoni (commented):
Looks good to me. I was actually noticing this bug this morning and wanted to make a PR but you beat me to it. 

Just doing some final tests and then I will approve and merge this in.

korenyoni (approved):

PR #171, opened 2021-06-10
add allow_ssl_requests_only to force ssl communication

mcalhoun:
## what

* Add the `allow_ssl_requests_only` flag to force the s3 bucket to only communicate over ssl

## why

* AWS Foundational Security Best Practices

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -608,3 +608,9 @@ variable "realtime_log_config_arn" {
   default     = null
   description = "The ARN of the real-time log configuration that is attached to this cache behavior"
 }
+
+variable "allow_ssl_requests_only" {
+  type        = bool
+  default     = true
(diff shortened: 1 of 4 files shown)


variables.tf
```diff
@@ -608,3 +608,9 @@ variable "realtime_log_config_arn" {
   default     = null
   description = "The ARN of the real-time log configuration that is attached to this cache behavior"
 }
+
+variable "allow_ssl_requests_only" {
+  type        = bool
+  default     = true
```
nitrocode, line 614: Can defaulting this to `true` break any current deployments ?

mcalhoun: it shouldn't...since this module is specifically to create the S3 bucket that is the origin for CloudFront, and CloudFront uses SSL by default.

nitrocode (approved):

PR #173, opened 2021-06-30
Enable compression of responses by default

alexjurkiewicz:
Compression of responses is free, and should always be a net-positive.


nitrocode (approved):

alexjurkiewicz:
Looks like the test failure is not this PR's fault

PR #176, opened 2021-07-07
Add trusted_* vars per ordered_cache

martinlindner:
## what
Add individual `trusted_signers`, `trusted_key_groups` attributes to `var.ordered_cache`.

## why
* "Private content" / signature checks should be configurable on a per-behavior level instead of just enforcing the global settings for all.


martinlindner:
Quick ping @Makeshift, @brcnblc. Sorry to bug you. Any chance to get this reviewed? 

I actually have a prod use case for per-behavior `trusted_key_groups`, and it'd be awesome if I could use this module directly instead of my own fork.

nitrocode (approved):

PR #177, opened 2021-07-21
Add var.create_origin_bucket

alexjurkiewicz:
This adjustment allows for a slight improvement in functionality -- it's
now possible to have this module create the origin bucket, but manually
specify the name to use.

This change works around a problem where it's not possible to use a
computed value for the origin bucket value. For example, if you create
an S3 bucket using the Cloudposse S3 bucket module, you'll see an error
like:

> Error: Invalid count argument
>
>   on .terraform/modules/my_cdn/main.tf line 312, in data "aws_s3_bucket" "origin":
>  312:   count  = local.enabled && (var.origin_bucket != null) ? 1 : 0
>
> The "count" value depends on resource attributes that cannot be determined
> until apply, so Terraform cannot predict how many instances will be created.
> To work around this, use the -target argument to first apply only the
> resources that the count depends on.

Unfortunately this fix just moves the corner case, rather than
eliminating it. If a user specifies an existing bucket for
var.origin_bucket, they must set create_origin_bucket to false or a
spurious bucket will be created.

---

## what
I ran into this problem while deploying a simple CDN as follows:

```terraform
module "label" {
  source    = "git::https://github.com/cloudposse/terraform-null-label.git?ref=0.24.1"
  # ...
}
module "origin_bucket" {
  source  = "cloudposse/s3-bucket/aws"
  context = module.label.context
  # ...
}
module "cdn" {
  context = module.label.context
  origin_bucket = module.origin_bucket.bucket_id
  # ...
}
```

## why
* It's sad Terraform has no ability to completely squash this corner case. But it's good to eliminate the possibility of accidentally building non-reproducible Terraform configurations


nitrocode:
just wondering if you set `origin_bucket = module.label.id` and set a `depends_on`, would this work without this pr change?

```hcl
module "cdn" {
  source        = "cloudposse/cloudfront-s3-cdn/aws"
  context       = module.label.context
  origin_bucket = module.label.id
  # ...
  depends_on = [module.origin_bucket]
}
```

sebastianmacarescu:
I also think this is very useful in case we want to override the origin bucket name while still allowing this module to create it.

PR #179, opened 2021-08-06
Disable SSL-only policy if ACM certificate given

### LINKED ISSUE #175: 403 AccessDenied for static website
## Describe the Bug
Using the module with 
```
  dns_alias_enabled       = true
  website_enabled         = true
  versioning_enabled      = true
  allow_ssl_requests_only = true
```

Resources created OK, but I get access denied on every resource on the website (with or without cloudfront)

## Expected Behavior
Should not happen if we access the website from CloudFront URL.

## Steps to Reproduce
```
module "cdn" {
  source = "cloudposse/cloudfront-s3-cdn/aws"

  version                   = "0.73.0"
  enabled                   = var.enabled
  aliases                   = ["${var.name}.${var.parent_zone_name}"] // TODO: strip "production" from cnames
  dns_alias_enabled         = true
  website_enabled           = true
  versioning_enabled        = true
  parent_zone_id            = var.parent_zone_id
  namespace                 = var.namespace
  environment               = var.environment
  name                      = var.name
  allow_ssl_requests_only   = true # NOTE: breaking access from cloudfront, maybe needs redirection to HTTPS
  deployment_principal_arns = {}

  minimum_protocol_version = "TLSv1"
  acm_certificate_arn      = var.acm_certificate_arn
}
```

Working around it is just disabling SSL (allow_ssl_requests_only = false) 

Cheers :)

emyller:
## what
* Disables SSL-only requests restriction at S3 static-website settings when Cloudfront will handle TLS given an ACM certificate.

## why
* There is a conflict when setting up both CDN-level TLS and S3-level SSL, which makes S3 respond with a HTTP 403 error.

## references
* Closes #175

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -184,7 +184,7 @@ data "aws_iam_policy_document" "deployment" {
 }
 
 data "aws_iam_policy_document" "s3_ssl_only" {
-  count = var.allow_ssl_requests_only ? 1 : 0
+  count = var.allow_ssl_requests_only && var.acm_certificate_arn == "" ? 1 : 0
   statement {
     sid     = "ForceSSLOnlyAccess"
     effect  = "Deny"


emyller:
To maintainers: can you please explain a use case justifying `allow_ssl_requests_only=true` since this module includes a Cloudfront CDN with TLS settings of its own? Thanks!

osterman:
When deploying AWS Config and SecurityHub with the AWS Well-Architected Best Practices, it will raise an issue if TLS on a bucket is not mandatory.

osterman:
If the module has problems with this enabled, then I suspect we have a cloud front misconfiguration for the origin, since I can't see why it should ever send non TLS requests.

emyller:
> If the module has problems with this enabled, then I suspect we have a cloud front misconfiguration for the origin, since I can't see why it should ever send non TLS requests.

Gotcha. My guess is that Cloudfront would communicate _internally_ with S3, similarly to an ALB when it communicates with the hosts within a target group -- both scenarios not needing TLS. I can see a flaw in the Cloudfront-S3 scenario though: in case someone has the bucket name, they could make requests without TLS.

I can help investigating why the origin fails making TLS requests to S3 in a case like #175. I doesn't happen everywhere -- I've used this module before and didn't see this issue.

dawilk:
> > If the module has problems with this enabled, then I suspect we have a cloud front misconfiguration for the origin, since I can't see why it should ever send non TLS requests.
> 
> Gotcha. My guess is that Cloudfront would communicate _internally_ with S3, similarly to an ALB when it communicates with the hosts within a target group -- both scenarios not needing TLS. I can see a flaw in the Cloudfront-S3 scenario though: in case someone has the bucket name, they could make requests without TLS.
> 
> I can help investigating why the origin fails making TLS requests to S3 in a case like #175. I doesn't happen everywhere -- I've used this module before and didn't see this issue.

I believe the issue is with the default origin protocol policy here: https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/blob/master/main.tf#L413
the origin generated by the default s3_origins var will have this problem - it gets forced to http-only. I think a fix would be to have the default set to `match-viewer` (the AWS default) and then let allow_ssl_requests_only=true change it to https-only.
I'd even be in favor of keeping allow_ssl_requests_only=true as default but make sure the origin_protocol_policy on the resource is set appropriately to align with the bucket policy.

PR #169, opened 2021-05-26
Add support for a custom cache policy for ordered cache

linhkikuchi:
## what
* To add support for a custom cache policy for ordered cache

PR #188, opened 2021-10-19
Fix: ordered_cache_behavior forwarded_values dynamic block

neilbartley:
## what
* Corrects logic for `ordered_cache_behavior -> forwarded_values`
* Should be `If a cache policy or origin request policy is specified, we cannot include a 'forwarded_values' block at all in the API request`

## why
* If I only specify one of `cache_policy_id` or `origin_request_policy_id` then the `forwarded_values` block would still be added, triggering an API error:
> `Error: error updating CloudFront Distribution (XXXXXXXXXXXXXX): InvalidArgument: The parameter ForwardedValues cannot be used when a cache policy is associated to the cache behavior.`
* Pre PR functionality:

| `cache_policy_id` supplied? | `origin_request_policy_id` supplied? | `forwarded_values` block generated? | correct behaviour? |
| :--- | :---: | :---: | :---: |
| ✅ | ✅  | ❌ | ✅ |
| ✅ | ❌  | ✅ | ❌ |
| ❌ | ✅  | ✅ | ❌ |
| ❌ | ❌  | ✅ | ✅ |

## references
* Added in #169


max-lobur:
Thanks for this @neilbartley !

Do you think it might be easier if we 
- add an explicit boolean (on/off) to forwarded_values 
- also make a forwarded_values group of variables a separate object

korenyoni:
Added a `needs-discussion` label while awaiting a response on https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/188#issuecomment-989721183

jabcoder:
Yeah, I think I am running into a similar issue. This module has been a lifesaver and for what it's worth it's the only issue I have run into so far.  While creating ordered_cache it is hard to tell what depends on what (by aws design it seems) and then it makes specify a lambda function arn even though I don't want to lol. Although I may just be doing something incorrect.

neilbartley:
Sorry, this fell off my radar. Looks like @korenyoni fixed in #210 🙌

Issue #189, opened 2021-10-26
Error if Origin Bucket has MFA Delete Enabled


sgtoj:
Found a bug? Maybe our [Slack Community](https://slack.cloudposse.com) can help. 

[![Slack Community](https://slack.cloudposse.com/badge.svg)](https://slack.cloudposse.com)

## Describe the Bug
If MFA Delete has been enabled on the origin bucket, this module attempts to set `mfa_delete` to `false`. Doing so causes an error:

```
AccessDenied: Mfa Authentication must be used for this request
```

## Expected Behavior
Ignore changes to MFA delete configuration.  

## Steps to Reproduce
1. deploy this module
2. enable mfa delete on the origin bucket
3. make changes to the module's configuration to trigger deployment on next deployment
4. deploy changes
5. See error

## Screenshots
If applicable, add screenshots or logs to help explain your problem.

## Environment (please complete the following information):
- terraform v1.0.9
- terraform aws provider v3.63.0

## Additional Context
N/A

Nuru:
We are not going to support MFA delete on See buckets. See details [here](https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/190#issuecomment-1050796576).

PR #190, opened 2021-10-26
ignore origin bucket mfa delete

sgtoj:
## what
* ignore `mfa_delete` property for origin s3 bucket

## why
* `mfa_delete` is not controlled by this module

## references
* https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/issues/189

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -4,7 +4,7 @@ locals {
   # Encapsulate logic here so that it is not lost/scattered among the configuration
   website_enabled           = local.enabled && var.website_enabled
   website_password_enabled  = local.website_enabled && var.s3_website_password_enabled
-  s3_origin_enabled         = local.enabled && !var.website_enabled
+  s3_origin_enabled         = local.enabled && ! var.website_enabled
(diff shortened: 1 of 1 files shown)


Nuru:
We have made the decision not to support S3 buckets that have MFA delete enabled. It was annoying with AWS provider v3 when the `mfa_delete` did not actually enable or disable the feature, it simply tracked its state. However, AWS provider v4 is now out and not only does that field set the state, when MFA delete is set, you have to supply an MFA token any time you want Terraform to work with the bucket. This is completely at odds with automation, and while there can be a good reason to enable MFA delete, our professional opinion is that in those cases, you should not be automating the provisioning and configuration of the S3 bucket with Terraform. Therefore, we are not going to support buckets where MFA delete is enabled.

You may want to use our companion module, [terraform-aws-cloudfront-cdn](https://github.com/cloudposse/terraform-aws-cloudfront-cdn), to take over the Cloudfront configuration and plug in your manually configured S3 bucket to use with it.

Closes #190 (this PR) and closes #189.

Issue #192, opened 2021-11-08
Add support for CloudFront Response Headers Policy


itschrishudson:
## Describe the Feature

AWS CloudFront now supports custom response headers natively without having to set these at origin or set via Lambda.

https://aws.amazon.com/about-aws/whats-new/2021/11/amazon-cloudfront-supports-cors-security-custom-http-response-headers/

This was added to the Hashicorp AWS Provider in 3.64.0
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_response_headers_policy

Please add support for these new data objects within the S3-CDN module (feature request also created for `terraform-aws-cloudfront-cdn`)

## Expected Behavior

When specified the module will allow passing the appropriate config maps to create request header policies in CloudFront.

## Use Case

Users of AWS CloudFront may require setting of request headers for security, caching or other purposes. This has previously required these headers to be set at the origin, or an additional lambda to set these headers. AWS has now enabled this to be set as part of the CloudFront distribution configuration.

## Describe Ideal Solution
 
Additional map(s) similar to the `ordered_cache` attribute that allows the appropriate config for a CORS/custom or security header policy to be provided.

## Alternatives Considered

As described earlier, setting the request header at the origin, or creating an additional lambda. For some use cases both of these approaches may not be possible for various reasons.

PR #195, opened 2021-11-10
Add response headers policy input for CloudFront distributions default behaviour

itschrishudson:
## what
* Adds an input to pass a response headers policy ID into the CloudFront distribution in the default cache behaviour.

## why
* AWS have recently released the ability to generate response header policies to add response headers for CORS/Security and custom use cases instead of setting at the origin or using a Lambda.
* The current module does not allow you to pass in the ID of a policy in order to associate it with a the CloudFront distributions default behaviour.

## references
* https://aws.amazon.com/blogs/networking-and-content-delivery/amazon-cloudfront-introduces-response-headers-policies/
* https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution#response_headers_policy_id
* Part of solution to #192


variables.tf
```diff
@@ -175,6 +175,13 @@ variable "price_class" {
   description = "Price class for this distribution: `PriceClass_All`, `PriceClass_200`, `PriceClass_100`"
 }
 
+
```
mihaiplesa, line 178: extra newline not needed

itschrishudson: Fixed

variables.tf
```diff
@@ -175,6 +175,12 @@ variable "price_class" {
   description = "Price class for this distribution: `PriceClass_All`, `PriceClass_200`, `PriceClass_100`"
 }
 
+variable "response_headers_policy_id" {
+  type        = string
+  description = "The identifier for a response headers policy - optional"
```
nitrocode, line 180:
```suggestion
  description = "The identifier for a response headers policy"
```

nitrocode (approved):

nitrocode:
@itschrishudson thanks for the contribution. This has been released as https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/releases/tag/0.76.0 🎉

PR #198, opened 2021-11-15
Added support for forward_cookies_whitelisted_names

rafaljanicki:
## what
I've added support for whitelisted_names (https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution#whitelisted_names) if forward_cookies is set to "whitelist"

## why
Currently, the only option to forward cookies is to set it to "all" which isn't perfect

## references
* I've used naming and approach from another Cloudposse module - https://github.com/cloudposse/terraform-aws-cloudfront-cdn/blob/master/variables.tf


rafaljanicki:
Sorry for hassling @nitrocode , but it'd make my life so much easier if we could get it merged. Let me know if there's anything I should add here :)

nitrocode (approved):

nitrocode:
thanks @rafaljanicki for the contribution!

PR #201, opened 2021-12-13
Chore: Bump Dependency Module Versions

korenyoni:
## what

- Bump `route53-alias` module version to latest.
- Bump `s3-log-storage` module to latest.

## why

- Dependency modules do not currently support `tenant` label (they are not at the version which supports it).
- General module hygiene: keep dependency modules up to date.

## references
- https://github.com/cloudposse/terraform-null-label/releases/tag/0.25.0
- Supersedes and closes #187

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -15,8 +15,8 @@
 
 # Cloud Posse must review any changes to standard context definition,
 # but some changes can be rubber-stamped.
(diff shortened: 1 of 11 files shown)


Nuru (approved):
Approved as-is, but consider incorporating some other open PRs and fixing some other open issues in this same, on PR, so we get more fixed in the one release.

max-lobur (approved):

max-lobur (approved):

PR #204, opened 2021-12-20
Feat: `lambda@edge` submodule

korenyoni:
## what
* Add submodule for creating Lambda@Edge functions
* Add test for Lambda@Edge submodule
* Ensure all tests run in parallel.
* Fix `versions.tf` in `examples/complete` according to TFLint

## why
* Lambda@Edge are a potentially common use case for this repo, and a submodule for creating Lambda@Edge functions for use with the root module should be created. 

## references
* N/A

## notes

Please note that a side effect of this change is that the test suite will now take ~30 minutes to run.


modules/lambda@edge/package.tf
```diff
@@ -0,0 +1,10 @@
+data "archive_file" "lambda_zip" {
+  for_each = local.functions
+
+  type             = "zip"
+  source_dir       = each.value.local_path
+  output_file_mode = "0666"
+  output_path      = "${path.module}/archives/${each.key}.zip"
```
osterman, line 7: Can we also support inline files?

korenyoni: Sure, i'll look into that. I don't think there's a reason we can't.

modules/lambda@edge/outputs.tf
```diff
@@ -0,0 +1,10 @@
+output "lambda_function_association" {
+  description = "TODO"
```
osterman, line 2: FYI

modules/lambda@edge/package.tf
korenyoni: I've updated the schema.

modules/lambda@edge/outputs.tf
korenyoni: Created a description.

goruha (dismissed):

goruha (approved):

goruha (dismissed):

modules/lambda@edge/main.tf
```diff
@@ -0,0 +1,94 @@
+locals {
+  enabled                   = module.this.enabled
+  destruction_delay_enabled = local.enabled && try(length(var.destruction_delay), 0) > 0
+  functions                 = local.enabled ? var.functions : {}
```
aknysh, line 4: did you try to plan the module with `local.enabled=false`?
Terraform usually complains that both sides of the `:` operator must have the same types and the same item types.
If TF does that in this case, then the solution can be:

```
functions  = {for k,v in var.functions : k => v if local.enabled }
```

modules/lambda@edge/main.tf
```diff
@@ -0,0 +1,94 @@
...
+}
+
+module "role" {
+  for_each = local.functions
+
+  source  = "cloudposse/iam-role/aws"
+  version = "0.14.0"
+
+  use_fullname       = true
+  policy_description = "Allow ${module.function_label[each.key].id} Write CloudWatch Logs"
```
aknysh, line 54:
```suggestion
  policy_description = "Allow ${module.function_label[each.key].id} write to CloudWatch Logs"
```

modules/lambda@edge/main.tf
```diff
@@ -0,0 +1,94 @@
...
+  role_description   = "IAM role for ${module.function_label[each.key].id}"
```
aknysh, line 55:
```suggestion
  role_description   = "IAM role for ${module.function_label[each.key].id} Lambda function"
```

aknysh (commented):
LGTM, a few nitpicks

modules/lambda@edge/main.tf
korenyoni: @aknysh yes, the test suite has a few cases where `enabled=false`, and there wasn't an issue.

aknysh (approved):

PR #207, opened 2022-01-11
feat: add origin-shield

syphernl:
## what
* Add variables to enable the Origin Shield for the CloudFront distribution

## why
* Using Origin Shield can help reduce the load on your origin.

## references
* https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution#origin_shield
* Also implemented recently in other CF module: https://github.com/cloudposse/terraform-aws-cloudfront-cdn/pull/78


syphernl:
✅  Verified on my own stack that it works.

The automated tests however will likely only work once https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/208 has been merged since the AWS Provider is now explicitly checking for a value to be set for `ssl_support_method`.

variables.tf
```diff
@@ -223,6 +223,18 @@ variable "origin_request_policy_id" {
...
 }
 
+variable "origin_shield" {
+  type = object({
+    enabled = bool
+    region  = string
+  })
+  description = <<-EOT
+    The CloudFront Origin Shield settings.
+    If `region` is set to `auto` it will determine the closest region where Origin Shield is available.
```
korenyoni, line 233: This is really clever, but I've never seen this type of behaviour in any of our modules. I'm not sure if maybe this is an exception, but I definitely am going to need someone else's opinion on it.

In my opinion, it's a little too "smart" for a Terraform configuration.

main.tf
```diff
@@ -74,6 +74,34 @@ locals {
...
+    # Regions where Origin Shield is NOT available (choose closest region)
+    us-west-1    = "us-west-2"      # US West (N. California)
+    af-south-1   = "eu-west-1"      # Africa (Cape Town)
+    ap-east-1    = "ap-southeast-1" # Asia Pacific (Hong Kong)
+    ca-central-1 = "us-east-1"      # Canada (Central)
+    eu-south-1   = "eu-central-1"   # Europe (Milan)
+    eu-west-3    = "eu-west-2"      # Europe (Paris)
+    eu-north-1   = "eu-west-2"      # Europe (Stockholm)
+    me-south-1   = "ap-south-1"     # Middle East (Bahrain)
+  }
```
korenyoni, line 104: See my other comment on this type of behaviour not having much precedent in our modules.

variables.tf
```diff
@@ -434,6 +446,10 @@ variable "custom_origins" {
       origin_keepalive_timeout = number
       origin_read_timeout      = number
     })
+    origin_shield = object({
+      enabled = bool
+      region  = string
+    })
```
korenyoni, line 452: It sucks that we have to make such a complex object even more complex. It's a breaking change, but there really isn't any other logical place for it IMO.

If we didn't have to opt in it'd be cool (https://www.terraform.io/language/functions/defaults)

But it is what it is.

variables.tf
```diff
@@ -446,6 +462,10 @@ variable "s3_origins" {
     domain_name = string
     origin_id   = string
     origin_path = string
+    origin_shield = object({
+      enabled = bool
+      region  = string
+    })
```
korenyoni, line 468: See comment on breaking change

korenyoni (commented):
In general this LGTM, with the exception that I think that being able to set `region = "auto"` and the map that comes alongside it is a little too "smart" for Terraform. This is more something that should happen on the provider side. I'd like someone else to weigh in. @osterman ?

It'd also be nice to get @Nuru 's opinion on the addition of an `origin_shield` attribute to `s3_origins` and `custom_origins`.

korenyoni:
So @syphernl looks good, not requesting any changes but want to get @osterman and @Nuru's opinions on what I've identified.

main.tf
osterman: No, we do similar things here: https://github.com/cloudposse/terraform-aws-utils/blob/master/main.tf

variables.tf
osterman: I think it's acceptable, especially if it doesn't change to often and it's easy to add other regions as they come online.  

My main suggestion though is we should move the `local.origin_shield_region_mapping` to a `var. origin_shield_region_mapping` with a default of this map so that anyone can change the behavior without forking the module.

main.tf
osterman: ironically, I see we are using a `local` for it too, but I think a `var` would be more appropriate in this case.

korenyoni: @osterman right, but I didn't think https://github.com/cloudposse/terraform-aws-utils/blob/master/main.tf of a proper example since it's a utility module specifically meant to map region short-hands and not a regular module. But anyways, if it's acceptable precedent in your view, then that helps.

variables.tf
syphernl: Good suggestion. I have changed the `local` into a `var` so anyone can override its behavior.

variables.tf
syphernl: I agree. Let's hope the `defaults` function will soon move out of the experimental phase.

syphernl:
@korenyoni I have changed the local into a var as @osterman suggested.

korenyoni:
Thanks @syphernl —

@osterman please take another look. Deferring approval to you.

variables.tf
osterman: @syphernl how much of a blocker is this for you? We're trying to keep breaking changes to a minimum, and yes, desperate for `defaults` function to graduate from experimental.

syphernl: @osterman Our customers would like to have this implemented in their environment. 
For the time being we could probably use our own fork/branch instead although from update-perspective that is not ideal.

korenyoni:
Hi @syphernl —

I've spoken to @Nuru who needs more time to be able to get a good look at this.

korenyoni:
> @korenyoni @Nuru Do you have any updates regarding this? Our project development is nearing completion and we'd like to enable the Origin Shield for this soon.

At the moment it's in @Nuru 's hands. 

From @Nuru :

>  I had to put this on hold because of the AWS 4.0 upgrade.

alexjurkiewicz:
I implemented the non-breaking-change portion of this in #247

PR #208, opened 2022-01-11
fix: set viewer_certificate.ssl_support_method to a non-empty value

syphernl:
## what
* Sets the `viewer_certificate.ssl_support_method` to a non-empty value

## why
* Since [AWS Provider 3.71.0](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v3.71.0) the `viewer_certificate.ssl_support_method` is being validated on plan-time and can no longer be an empty string but must be `sni-only` or `vip` (of which the latter incurs costs)

## references
* https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution#ssl_support_method

### DIFF
--- a/README.md
+++ b/README.md
@@ -692,7 +692,7 @@ In general, PRs are welcome. We follow the typical "fork-and-pull" Git workflow.
 
 ## Copyright
 
-Copyright © 2017-2021 [Cloud Posse, LLC](https://cpco.io/copyright)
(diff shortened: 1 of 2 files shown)


korenyoni (dismissed):

korenyoni:
Going to merge with BC errors because I have them fixed in #209

korenyoni (approved):

PR #209, opened 2022-01-12
Feat: Add Support for S3 Object Ownership Controls

korenyoni:
## what
- Add support for S3 object ownership controls.
- Misc: add BridgeCrew exceptions.

## why

S3 object ownership controls should be configurable for the origin bucket behind the CloudFront distribution. If the origin bucket is used within CI pipelines, cross-account writes to the bucket will result in the object being owned by that account. This can result in 403's when attempting to access content in the CF distribution, unless the S3 object ownership setting is set to `BucketOwnerEnforced`.

> By default, when another AWS account uploads an object to your S3 bucket, that account (the object writer) owns the object, has access to it, and can grant other users access to it through ACLs.

The new S3 object ownership variable has as a default value that has the same effect if the new `s3_bucket_ownership_controls` were not to exist at all. Thus, this is a backwards-compatible release.

## references
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html
- https://github.com/cloudposse/terraform-aws-s3-bucket/pull/109

### DIFF
--- a/README.md
+++ b/README.md
@@ -438,6 +438,7 @@ Available targets:
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 0.13.0 |
 | <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 3.41.0 |
 | <a name="requirement_random"></a> [random](#requirement\_random) | >= 2.2 |
+| <a name="requirement_time"></a> [time](#requirement\_time) | >= 0.7 |
 
 ## Providers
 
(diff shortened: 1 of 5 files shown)


max-lobur (dismissed):

korenyoni:
Breaking because of new plan-time validation in https://github.com/hashicorp/terraform-provider-aws/releases/tag/v3.71.0

max-lobur (approved):

PR #210, opened 2022-01-21
[main.tf] Updated conditions with included cache_policy_id variable

eboboshka:
## what
* Fixed conditions when using cache_policy_id

## why
* When using `cache_policy_id`, the module tries to change ttl values from zero to the default values.
```
      ~ default_cache_behavior {
          ~ default_ttl            = 0 -> 60
          ~ max_ttl                = 0 -> 31536000
            # (10 unchanged attributes hidden)

            # (1 unchanged block hidden)
        }
```
* There is no need to use the mandatory ARN lambda in Cusom Behavior.

## references
* https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/188


korenyoni (approved):
LGTM.

PR #211, opened 2022-01-21
Added new response_headers_policy_id argument for a block with Custom behavior

eboboshka:
[main.tf, variables.tf] 

## what
* Added new response_headers_policy_id argument for a block with Custom behavior

## why
* They just forgot about the cusom block `ordered_cache_behavior`

## references
* https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/issues/192

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -549,10 +549,11 @@ resource "aws_cloudfront_distribution" "default" {
         }
       }
 
-      viewer_protocol_policy = ordered_cache_behavior.value.viewer_protocol_policy
-      default_ttl            = ordered_cache_behavior.value.default_ttl
-      min_ttl                = ordered_cache_behavior.value.min_ttl
-      max_ttl                = ordered_cache_behavior.value.max_ttl
+      viewer_protocol_policy     = ordered_cache_behavior.value.viewer_protocol_policy
+      default_ttl                = ordered_cache_behavior.value.default_ttl
+      min_ttl                    = ordered_cache_behavior.value.min_ttl
+      max_ttl                    = ordered_cache_behavior.value.max_ttl
+      response_headers_policy_id = try(ordered_cache_behavior.value.response_headers_policy_id, null)
 
       dynamic "lambda_function_association" {
         for_each = ordered_cache_behavior.value.lambda_function_association
--- a/variables.tf
+++ b/variables.tf
@@ -388,10 +388,11 @@ variable "ordered_cache" {
     cache_policy_id          = string
     origin_request_policy_id = string
 
-    viewer_protocol_policy = string
-    min_ttl                = number
-    default_ttl            = number
-    max_ttl                = number
+    viewer_protocol_policy     = string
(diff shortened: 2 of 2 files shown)


reda-sj:
Just what I was looking for. nice work

main.tf
```diff
@@ -549,10 +549,11 @@ resource "aws_cloudfront_distribution" "default" {
...
 
-      viewer_protocol_policy = ordered_cache_behavior.value.viewer_protocol_policy
-      default_ttl            = ordered_cache_behavior.value.default_ttl
-      min_ttl                = ordered_cache_behavior.value.min_ttl
-      max_ttl                = ordered_cache_behavior.value.max_ttl
+      viewer_protocol_policy     = ordered_cache_behavior.value.viewer_protocol_policy
+      default_ttl                = ordered_cache_behavior.value.default_ttl
+      min_ttl                    = ordered_cache_behavior.value.min_ttl
+      max_ttl                    = ordered_cache_behavior.value.max_ttl
+      response_headers_policy_id = try(ordered_cache_behavior.value.response_headers_policy_id, null)
```
Gowiem, line 556: You don't need the `try` here. Because `var. ordered_cache` is an object, all of the attributes specified in the type need to be passed. So this will always have a value of at least `null` or `""` from the module consumer. Mind removing as it may confuse other contributors in the future?

Gowiem (changes requested):

main.tf
eboboshka: I would like to keep the try function because there is no optional function in Terraform variables, which means that for custom behavior you have to specify this parameter, which uselessly increases the amount of code
https://github.com/hashicorp/terraform/issues/19898

eboboshka: The `ordered_cache` variable is also declared without default values: https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/blob/master/variables.tf#L412

Gowiem: @eboboshka Yeah, so for both of those points that means that object's value will have to provide a value for the attribute when it is passed from the root module, regardless of the `try` or not. `tf validate` (and the various commands that run `validate` as a pre-step) will not allow a consumer to pass a value to `ordered_cache` without it including `response_headers_policy_id`. So the `try` becomes redundant as it will always be given a value since there is no option to do otherwise.

eboboshka: You're probably right. I'm just trying to remove unnecessary options since I'm using your module as a for_each module.
Anyway, if you think it''s necessary to remove `try`, you can change my commit

Just as an example
image
image

Gowiem (approved):

main.tf
Gowiem: Thanks for the update @eboboshka!

Gowiem:
@eboboshka Thanks for the contribution! Released as https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/releases/tag/0.82.0.

PR #213, opened 2022-01-25
fix: handle viewer_certificate.ssl_support_method with CF default certificate

syphernl:
## what
* Sets `ssl_support_method` to `null` when default certificate is being used instead of setting it to `sni-only` in all cases.

## why
* Setting this to an explicit value (`sni-only`) in conjunction with the "default certificate" results in the state not being idempotent. 

## references

* https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/208


main.tf
```diff
@@ -463,7 +463,7 @@ resource "aws_cloudfront_distribution" "default" {
 
   viewer_certificate {
     acm_certificate_arn            = var.acm_certificate_arn
-    ssl_support_method             = "sni-only"
+    ssl_support_method             = local.use_default_acm_certificate ? null : "sni-only"
```
korenyoni, line 466: https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/208

Does `null` behave differently than `""` ? in terms of how the provider validates this field?

korenyoni: Right I believe because it's an optional value it'll be fine, but `""` is subject to the validation method https://github.com/hashicorp/terraform-provider-aws/blob/c47375e5482213883df51795322e2e2b55bc9348/internal/service/cloudfront/distribution.go#L694

korenyoni (dismissed):
LGTM.

PR #218, opened 2022-02-15
Update versions.tf

rusowyler:
## what
* Restrict AWS provider version

## why
* \>= 3.64 because response_headers_policy_id first appeared on version: 3.64.0 (November 04, 2021)
* < 4.0.0 many breaking changes affect this module.

## references
* https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md#3640-november-04-2021
* https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md#400-february-10-2022

### DIFF
--- a/README.md
+++ b/README.md
@@ -436,7 +436,7 @@ Available targets:
 | Name | Version |
 |------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 0.13.0 |
-| <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 3.41.0 |
+| <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 3.64.0, < 4.0.0 |
 | <a name="requirement_random"></a> [random](#requirement\_random) | >= 2.2 |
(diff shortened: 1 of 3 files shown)


Gowiem (approved):

Nuru:
@Gowiem We do not allow provider pinning except for `>=`. This PR should not have passed automated testing (I'm looking into that) and should not have been approved. 

Our policy is that if a new provider version breaks an existing module, users of that module should put the provider pin `e.g. < 4.0` in their component or use the `.terraform.lock.hcl` file. When the module is upgraded to work with the new provider, then it can be pinned, in this case `>= 4.0.0`.

Gowiem:
@Nuru gotcha and apologies, I didn't know about that. [Let's discuss in Slack here if you don't mind](https://sweetops.slack.com/archives/G014YEKDH4K/p1645465125550889?thread_ts=1645370067.250759&cid=G014YEKDH4K).

apanzerj:
Hi!

What's the outcome from that slack discussion? 

Is there plans to upgrade to 4.0?

diestrin:
It doesn't let me run it with the v4 provider still :(

PR #232, opened 2022-05-18
Set versioning_enabled default to false

nitrocode:
## what
* Set versioning_enabled default to false

## why
* No reason by default to version s3 static assets in a cloudfront s3 website

## references
* Implemented by this PR https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/121 probably to pass the bridgecrew check. The bridgecrew check in this case should be ignored.


Gowiem (approved):
This is a hard one as I feel like S3 is so cheap that versioning is something I always want on, but then I do get your point: it really shouldn't be needed for static CDN assets. 

You've got my approval, but I would get another review from someone internal on your end who might have a strong opinion on this matter before merging.

PR #234, opened 2022-06-23
Fix example to work

ezaurum:
## what
* Change example to work

## why
* It is not work because of alias invalid.

## references
* From error message, 
```
| Error: Invalid provider configuration alias
│ 
│ An alias must be a valid name. A name must start with a letter or underscore and may contain only letters, digits, underscores, and dashes.
```


joe-niland:
The changes don't seem to reflect the description. Feel free to update and we will reopen.

PR #236, opened 2022-08-15
Allow AWS provider version >= 4.9.0

nitrocode:
## what
- Allow AWS provider version `>= 4.9.0`

## references
- Version 4.0.0 had breaking changes, so this module had pinned AWS provider `< 4.0`. Breaking changes were replaced with deprecation warnings in v4.9.0, so we can now allow later versions.


Nuru (approved):
Changed to more appropriate version constraint

PR #238, opened 2022-08-23
Add http_version variable

kim0:


grifante:
It would be great to have this variable available

PR #241, opened 2022-08-29
Add var.create_cloudfront_origin_access_identity

### LINKED ISSUE #240: Using existing origin access identity fails with explicit context
Found a bug? Maybe our [Slack Community](https://slack.cloudposse.com) can help. 

[Slack Community](https://slack.cloudposse.com)

## Describe the Bug

Code:

```terraform
module "label" {
  source  = "cloudposse/label/null"
  version = "0.25.0"
  name        = "foo"
}
resource "aws_cloudfront_origin_access_identity" "this" {
  comment = module.label.id
}
module "cdn" {
  version = "0.82.5"
  # ...
  context = module.label.context
  cloudfront_origin_access_identity_iam_arn = aws_cloudfront_origin_access_identity.this.iam_arn
}
```

Gives error:
```
│ Error: Invalid count argument
│ 
│   on .terraform/modules/cdn/main.tf line 103, in resource "aws_cloudfront_origin_access_identity" "default":
│  103:   count = local.create_cloudfront_origin_access_identity ? 1 : 0
│ 
│ The "count" value depends on resource attributes that cannot be determined
│ until apply, so Terraform cannot predict how many instances will be
│ created. To work around this, use the -target argument to first apply only
│ the resources that the count depends on.
```

This local variable is defined as:

```terraform
locals {
  enabled = module.this.enabled
  create_cloudfront_origin_access_identity = local.enabled && length(compact([var.cloudfront_origin_access_identity_iam_arn])) == 0 # "" or null
}
```

I'm not 100% clear why this fails, but I believe it's because `module.this.enabled` can't be determined from static data.

## Environment (please complete the following information):

Anything that will help us triage the bug will help. Here are some ideas:
Terraform 1.2.6

alexjurkiewicz:
In some circumstances, it wasn't possible to automatically determine if
an Origin Access Identity should be created. Because of this, Terraform
requires us to always explicitly specify if the OAI should be created.

* Fixes #240


alexjurkiewicz:
This is working for me (in a private test)

PR #258, opened 2023-02-14
bugfix deprecate `deployment_principal_arns` (#257)

### LINKED ISSUE #257: Cannot create deployment principal and the CDN in the same terraform run
## Describe the Bug

I would like to create an IAM user and make it the deployment principal in the same terraform execution.

But this isn't possible due to the use of dynamic map keys.

This seems like a very simple and common use-case.

## Expected Behavior

I expect to be able to create my deployment principal and the CDN in one go.

## Steps to Reproduce
1. Run the tf module below

```terraform
resource "aws_iam_user" "deploy_user" {
  name = module.this.id
  tags = module.this.tags
}

module "cdn" {
  source                              = "cloudposse/cloudfront-s3-cdn/aws"
  version                             = "0.86.0"
  context                             = module.this.context
  cloudfront_access_logging_enabled   = true
  cloudfront_access_log_create_bucket = true
  deployment_principal_arns = {
    (aws_iam_user.deploy_user.arn) = [""]
    # this also doesn't work
    # "${aws_iam_user.deploy_user.arn}" = [""]
  }
}
```

**Result**:

```
╷
│ Error: Invalid for_each argument
│ 
│   on .terraform/modules/cdn/main.tf line 191, in data "aws_iam_policy_document" "deployment":
│  191:   for_each = local.enabled ? var.deployment_principal_arns : {}
│     ├────────────────
│     │ local.enabled is true
│     │ var.deployment_principal_arns is a map of list of string, known only after apply
│ 
│ The "for_each" map includes keys derived from resource attributes that cannot be determined until apply, and so Terraform cannot determine the full set of keys that will identify the instances of this resource.
│ 
│ When working with unknown values in for_each, it's better to define the map keys statically in your configuration and place apply-time results only in the map values.
│ 
│ Alternatively, you could use the -target planning option to first apply only the resources that the for_each value depends on, and then apply a second time to fully converge.
```

## Environment (please complete the following information):

Anything that will help us triage the bug will help. Here are some ideas:
 - TF version 1.3.8

## Additional Info

Hashicorp recommends storing only known values in map keys, and leaving all dynamic/unknown values in the map values ([source0](https://developer.hashicorp.com/terraform/language/meta-arguments/for_each#limitations-on-values-used-in-for_each), [source1](https://github.com/hashicorp/terraform/issues/30838#issuecomment-1096848884)).

Fixing it I think will require deprecating the existing `deployment_principal_arns` variable, and adding a new one that uses a map of objects:

```terraform
...
  deployment_principal_arns = {
    "deploy_user1" : {
      "arn" : aws_iam_user.deploy_user.arn,
      "path_prefix" : ""
    }
  }
...
```

abeluck:
## what
* This patch refactors the passing of deployment principals, such that it uses static/known map keys. 
* It replaces `deployment_principal_arns` with `deployment_principals`

## why

* This allows this module to be applied at the same time as the deployment principal (e.g., an iam user) is is deployed.
* Hashicorp recommends storing only known values in map keys, and leaving all dynamic/unknown values in the map values
([source0](https://developer.hashicorp.com/terraform/language/meta-arguments/for_each#limitations-on-values-used-in-for_each), [source1](https://github.com/hashicorp/terraform/issues/30838#issuecomment-1096848884)).

## references

* Use closes #257


deprecated.tf
```diff
@@ -7,6 +7,8 @@ locals {
   cloudfront_access_log_include_cookies = var.log_include_cookies == null ? var.cloudfront_access_log_include_cookies : var.log_include_cookies
   cloudfront_access_log_prefix          = var.log_prefix == null ? var.cloudfront_access_log_prefix : var.log_prefix
 
+  deployment_principals = var.deployment_principal_arns == null ? var.deployment_principals : { for arn, path_prefix in var.deployment_principal_arns : arn => { "arn" : arn, "path_prefix" : path_prefix } }
```
nitrocode, line 10: This logic returns the following error in our tests

```
        	            	Error: Inconsistent conditional result types
        	            	
        	            	  on ../../deprecated.tf line 10, in locals:
        	            	  10:   deployment_principals = var.deployment_principal_arns == null ? var.deployment_principals : { for arn, path_prefix in var.deployment_principal_arns : arn => { "arn" : arn, "path_prefix" : path_prefix } }
```

Could you format this to make the logic easier to read? Id also put the second portion of the ternary in a separate local.

```suggestion
  deployment_principal_arns = {
    for arn, path_prefix in var.deployment_principal_arns :
    arn => { "arn" : arn, "path_prefix" : path_prefix }
  }
  deployment_principals = var.deployment_principal_arns == null ? var.deployment_principals : local.deployment_principal_arns
```

Id also try to run the examples/complete module locally to make sure it works as expected. It will be faster iteration.

abeluck: sorry for the delay here.. @nitrocode I've pushed a patch that fixes the readability. I'll report back on running the `examples/complete` module.

abeluck: ..hmm we don't use route 53, so I don't have an R53 zones setup. Is there another way to run some tests?

joe-niland: @abeluck I re-ran the Terratest here and the result is the same as above

```
│ Error: Inconsistent conditional result types
│ 
│   on ../../deprecated.tf line 17, in locals:
│   17:   deployment_principals = var.deployment_principal_arns == null ? var.deployment_principals : local.deployment_principals_from_deprecated_deployment_principal_arns
│     ├────────────────
│     │ local.deployment_principals_from_deprecated_deployment_principal_arns is object with 2 attributes
│     │ var.deployment_principal_arns is map of list of string with 2 elements
│     │ var.deployment_principals is empty map of object
│ 
│ The true and false result expressions must have consistent types. The
│ 'true' value is map of object, but the 'false' value is object.
╵}
```

I believe the issues is because the [example](https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/blob/main/examples/complete/deployment.tf#L12) expects the map value to be a list of strings, whereas your variable is expecting a single prefix per arn.

abeluck: Thanks @joe-niland It was an easy fix. I pared down the complete test to be able to run it locally. It's working now both with the deprecated `var.deployment_principal_arns` and new `var.deployment_principals`.

I've updated the documentation in a few more places, and rebased the branch so the PR can be easily squash merged.

We could also change the complete example to use the new `var.deployment_principals`, what do you think?

joe-niland: Thanks @abeluck, that sounds good.
Yes, I think it makes sense to update the example.

joe-niland:
Hi @abeluck can you please run the following and commit the result?
Just need to update README.md

```
make init
make readme
```

abeluck:
> can you please run the following and commit the result?

Done!

deiga:
Looking forward to this getting merged!

deprecated.tf
joe-niland: @abeluck sorry for the delay - all looks good, but will you be updating the examples?

abeluck:
@joe-niland let me know if there is anything else I can do!

deprecated.tf
joe-niland: @abeluck see comment above re: examples :)

hans-d:
@joe-niland can you follow up on this one?

joe-niland:
@abeluck if you have a minute to update the main example to use `deployment_principals` then I think we're good to go. Sorry for the delay.

PR #264, opened 2023-03-29
Support overriding default origin ID

mchristopher:
## what

* Add support for `default_origin_id` variable, to support custom origin groups on the default behavior.
* Add optional `group_id` variable to origin_groups to avoid race conditions when specifying origin groups in behaviors.

## why

* There was no way to activate support for multi-region S3 buckets since origin groups have generated IDs, and you cannot access the `id` attributes outside of the module code.


Gowiem:
Hey @mchristopher, unfortunately, we need to run some automation to pass our tests. Mind doing the following locally, adding + committing the result, and pushing to your branch?

```
make init
make readme
```

Thanks!

Gowiem:
And actually it seems there is some lint errors in this repo overall, would you mind addressing those to move this PR forward?

PR #266, opened 2023-04-26
Fixed Deprecated aws_s3_bucket Options

afrazkhan:
The following options were deprecated, and have either been updated,
  or replaced with the new necessary resources:

  * `override_json` replaced with `override_policy_documents` option
  * `server_side_encryption_configuration` replaced with `aws_s3_bucket_server_side_encryption_configuration` resource
  * `versioning` replaced with `aws_s3_bucket_versioning` resource
  * `cors_rule` replaced with `aws_s3_bucket_cors_configuration` resource
  * `acl` replaced with `aws_s3_bucket_acl` resource


bshabae:
> The following options were deprecated, and have either been updated, or replaced with the new necessary resources:
> 
> * `override_json` replaced with `override_policy_documents` option
> * `server_side_encryption_configuration` replaced with `aws_s3_bucket_server_side_encryption_configuration` resource
> * `versioning` replaced with `aws_s3_bucket_versioning` resource
> * `cors_rule` replaced with `aws_s3_bucket_cors_configuration` resource
> * `acl` replaced with `aws_s3_bucket_acl` resource

Hi guys @dotCipher and @joe-niland!

Any updates on this? We are still seeing the following deprecation warning:

```
│ Warning: Argument is deprecated
│   on .terraform/modules/cloudfront_s3_site/main.tf line 248, in resource "aws_s3_bucket" "origin":
│  248: resource "aws_s3_bucket" "origin" {
│ 
│ Use the aws_s3_bucket_cors_configuration resource instead
```

davidjeddy:
Bump, checking in on the progress of this patch.

bshabae:
> Bump, checking in on the progress of this patch.

Thank you! We are looking forward to this update.

joe-niland:
Thanks @afrazkhan and sorry for the delay in reviewing.

There are some formatting issues. Could you please run the following to see the output?

```
make precommit/terraform
```

I believe they can be fixed by running `terraform fmt -recursive`

variables.tf
```diff
@@ -480,6 +480,17 @@ variable "versioning_enabled" {
   description = "When set to 'true' the s3 origin bucket will have versioning enabled"
 }
 
+variable "bucket_versioning" {
```
joe-niland, line 483: This effectively deprecates `versioning_enabled`. For backwards compatibility it should be mapped to the new variable with value "Enabled". 

It would also be good to remove its use from the examples.

joe-niland (changes requested):

variables.tf
afrazkhan: I might be misunderstanding, but since `var.versioning_enabled` is a `bool`, wouldn't I break backwards compatibility if I changed it to a string and set it's default to `Enabled`? Or do you mean to map it via something like doing some logic in `locals`?

Something like

```hcl
locals {
  bucket_versioning = var.versioning_enabled || var.bucket_versioning == "Enabled" ? "Enabled" : "Disabled"
}
```

and then use `local.bucket_versioning` everywhere instead?

PR #285, opened 2023-06-12
Update s3-log-storage to fix #267

soya-miyoshi:
## what

- I updated the s3-log-storage module to 1.4.0. 

## why

- fixs #267 
- The old version of this module fails to create a bucket because it tries to create a bucket whose ownership control is `BucketOwnerEnforced` (which is a new default configuration introduced by AWS) first with ACL.

PR #286, opened 2023-06-14
fix: ensure the static website configuration allows public access.

jhosteny:
## what

Addresses several issues:

```
│ Error: Error putting S3 policy: AccessDenied: Access Denied
│ 	status code: 403, request id: T4MJNDTVCB0SPQ91, host id: 6mdERYRKhtI8DpDBJgnu4dav3bWv3FmzIM+e4ibE5Jrf/71q3srieHDJuQY7w4FsMKqPM+7A9EaXUp2T6DG1Xw==
│ 
│   with module.cdn.aws_s3_bucket_policy.default[0],
│   on .terraform/modules/cdn/main.tf line 247, in resource "aws_s3_bucket_policy" "default":
│  247: resource "aws_s3_bucket_policy" "default" {
│ 
```

By inverting the dependency between the `aws_s3_bucket_policy.default` and the `aws_s3_bucket_public_access_block.origin` it avoids this error that I believe was introduced via the April S3 bucket changes.

```
│ Error: error creating CloudFront Distribution: InvalidArgument: The S3 bucket that you specified for CloudFront logs does not enable ACL access: tai-root-apex-logs.s3.amazonaws.com
│ 	status code: 400, request id: 510c7e09-f8c6-4225-b283-0cf5be42d587
│ 
│   with module.cdn.aws_cloudfront_distribution.default[0],
│   on .terraform/modules/cdn/main.tf line 372, in resource "aws_cloudfront_distribution" "default":
│  372: resource "aws_cloudfront_distribution" "default" {
```

To fix this, I made the `aws_cloudfront_distribution.default` resource depend on the existing `time_sleep.wait_for_aws_s3_bucket_settings`, so that the bucket policy and 

Finally, when the variable `website_enabled` is true, it changes the aws_s3_bucket_public_access_block to effectively make the bucket public. I am honestly not sure how this worked previously? I cannot get the bucket to work with redirects without this (https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn#your-critical-decision-s3-bucket-or-website).

Finally, this includes the change to the `s3-log module which is also present in other pending PRs.

## why

Ensure the module can apply cleanly without intervention, and have the correct public settings for `website_enabled`.

## references

- More complete fix than https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/282
- https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/285
- Should allow https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/pull/280/files to complete cleanly once this is merged (and that PR is rebased)

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -88,6 +88,10 @@ locals {
     "me-south-1"   = "ap-south-1"
   }
   origin_shield_region = local.enabled ? lookup(local.origin_shield_region_fallback_map, data.aws_region.current[0].name, data.aws_region.current[0].name) : "this string is never used"
+
+  # Public access is blocked if the explicit block setting is enabled, or if
+  # the bucket is not configured as a static website.
+  block_origin_public_access = var.block_origin_public_access_enabled || !var.website_enabled
 }
 
 ## Make up for deprecated template_file and lack of templatestring
@@ -245,6 +249,9 @@ resource "aws_s3_bucket_policy" "default" {
 
   bucket = local.origin_bucket.bucket
   policy = join("", data.aws_iam_policy_document.combined.*.json)
+
+  # Don't modify this bucket in two ways at the same time, S3 API will complain.
+  depends_on = [aws_s3_bucket_public_access_block.origin]
 }
 
 resource "aws_s3_bucket" "origin" {
@@ -307,15 +314,12 @@ resource "aws_s3_bucket" "origin" {
 }
 
 resource "aws_s3_bucket_public_access_block" "origin" {
-  count                   = (local.create_s3_origin_bucket || local.override_origin_bucket_policy) && var.block_origin_public_access_enabled ? 1 : 0
+  count                   = local.create_s3_origin_bucket || local.override_origin_bucket_policy ? 1 : 0
   bucket                  = local.bucket
-  block_public_acls       = true
-  block_public_policy     = true
-  ignore_public_acls      = true
-  restrict_public_buckets = true
-
-  # Don't modify this bucket in two ways at the same time, S3 API will complain.
-  depends_on = [aws_s3_bucket_policy.default]
+  block_public_acls       = local.block_origin_public_access
+  block_public_policy     = local.block_origin_public_access
+  ignore_public_acls      = local.block_origin_public_access
+  restrict_public_buckets = local.block_origin_public_access
 }
 
 resource "aws_s3_bucket_ownership_controls" "origin" {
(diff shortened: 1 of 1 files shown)


main.tf
```diff
@@ -245,6 +249,9 @@ resource "aws_s3_bucket_policy" "default" {
 
   bucket = local.origin_bucket.bucket
   policy = join("", data.aws_iam_policy_document.combined.*.json)
+
+  # Don't modify this bucket in two ways at the same time, S3 API will complain.
+  depends_on = [aws_s3_bucket_public_access_block.origin]
```
jhosteny, line 254: Inverts the dependency

main.tf
```diff
@@ -307,15 +314,12 @@ resource "aws_s3_bucket" "origin" {
 }
 
 resource "aws_s3_bucket_public_access_block" "origin" {
-  count                   = (local.create_s3_origin_bucket || local.override_origin_bucket_policy) && var.block_origin_public_access_enabled ? 1 : 0
+  count                   = local.create_s3_origin_bucket || local.override_origin_bucket_policy ? 1 : 0
```
jhosteny, line 317: Always create this resource, unless we didn't create the origin bucket and we are not overriding its policies.

*NOTE*: I did not test that case

main.tf
```diff
@@ -307,15 +314,12 @@ resource "aws_s3_bucket" "origin" {
...
-  block_public_policy     = true
-  ignore_public_acls      = true
-  restrict_public_buckets = true
-
-  # Don't modify this bucket in two ways at the same time, S3 API will complain.
-  depends_on = [aws_s3_bucket_policy.default]
+  block_public_acls       = local.block_origin_public_access
+  block_public_policy     = local.block_origin_public_access
+  ignore_public_acls      = local.block_origin_public_access
+  restrict_public_buckets = local.block_origin_public_access
```
jhosteny, line 322: Block unless it is a static website and we have not explicitly set the `block_origin_public_access_enabled` variable

main.tf
```diff
@@ -376,7 +380,7 @@ resource "aws_cloudfront_distribution" "default" {
   comment             = var.comment
   default_root_object = var.default_root_object
   price_class         = var.price_class
-  depends_on          = [aws_s3_bucket.origin]
+  depends_on          = [aws_s3_bucket.origin, time_sleep.wait_for_aws_s3_bucket_settings]
```
jhosteny, line 383: Wait for the eventually consistent settings so we don't have to re-apply

jhosteny:
Note that I tried this with both `website_enabled` set to true and false, and verified everything worked per my expectation. This should get a careful review. Specifically, I am not sure how the `acl = "private"` setting previously worked with the setting `website_enabled = true`. All of my origin buckets were set to be publicly readable, and I cannot see how the old version of the module that those were deployed with (`0.82.4`) would have set public access. I am assuming I changed that after the fact by hand.

PR #294, opened 2024-01-09
replace TLSv1.2_2019 with TLSv1.2_2021 as default policy

jamerply:
## what

This PR updates the `mimimum_protocol_version` variable so that it defaults to `TLSv1.2_2021` (the current recommended security policy recommended by AWS) instead of `TLSv1.2_2019`.

## why

The most current security policy is no longer `TLSv1.2_2019` but is `TLSv1.2_2021`.

## references

See the "Security Policy" heading under the ["Distribution Setting" section of the AWS CloudFront Documentation](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html#DownloadDistValuesGeneral) for further information.


Nuru (dismissed):
The problem is, this is a potentially breaking change that will manifest as some old clients failing to connect all of a sudden, a problem which will be difficult to trace back to updating this component. 

I think instead we should just recommend that users explicitly set this to the newer version.

jamerply:
@Nuru 
>The problem is, this is a potentially breaking change that will manifest as some old clients failing to connect all of a sudden, a problem which will be difficult to trace back to updating this component.

If users have pinned their modules to a particular version per CloudPosse's own recommendation, wouldn't that prevent the breaking change unless they knowingly updated to the newer version?

Gowiem:
@Nuru thoughts on just making this a major revision change? I agree that we should upgrade this for future consumers of this module, otherwise we're encouraging an old and outdated TLS. The other route is to remove the default altogether, but either way we'll want to do a major version rev. Let me know your thoughts and I can work with @jamerply to push this forward.

oycyc (commented):
Just came across this, think this would be good to have as default as it's officially recommended both on the docs (as the PR noted) and on the AWS console. I think a major bump would make sense in terms of @Nuru's concern about the potentially breaking change and backwards compatbility
image

Gowiem:
@oycyc agreed. I'm taking this internally to the Cloud Posse folks and asking for another review so that we move this forward. Thanks for reviving this PR with your comment 👍 

Sadly, we also have a test failure here: 

```
Run actions/setup-go@v5
/usr/bin/docker exec  e7aa506c37134f9795e46a4b5a510bca1ca734535914bf1bc0524a7ae95bfa85 sh -c "cat /etc/*release | grep ^ID"
Setup go version spec 1.20
Attempting to download 1.20...
matching 1.20...
Acquiring 1.20.14 from https://github.com/actions/go-versions/releases/download/1.20.14-7809684997/go-1.20.14-linux-x64.tar.gz
Extracting Go...
/usr/bin/tar xz --warning=no-unknown-keyword --overwrite -C /__w/_temp/5bc893cc-6754-4680-98e6-09dba2742765 -f /__w/_temp/248d[17](https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/actions/runs/14887918312/job/41812431151#step:13:18)bd-d695-4470-9221-f2c3548b1ccb
Successfully extracted go to /__w/_temp/5bc893cc-6754-4680-98e6-09dba2742765
Adding to the cache ...
Successfully cached go to /__t/go/1.[20](https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/actions/runs/14887918312/job/41812431151#step:13:21).14/x64
Added go to the path
Successfully set up Go version 1.20
Error loading shared library libresolv.so.2: No such file or directory (needed by /__t/go/1.20.14/x64/bin/go)
Error: Command failed: go env GOPATH
Error loading shared library libresolv.so.2: No such file or directory (needed by /__t/go/1.20.14/x64/bin/go)
```

I'll ask if we've seen that before.

osterman (approved):
This change is fine by me. It's non destructive, easily reversible by setting the variable, and a more secure default that we should optimize for.

Gowiem:
@Nuru @osterman to thread the needle here, I wonder if we should release a v1.0 as it stands today without this change and then follow up quickly with a v2.0 that includes this change. That will address @Nuru's comments surrounding automatic upgrades potentially breaking consumer usages of this module since once we're on v1.0, if they're upgrading to v2.0 then they should expect the incompatibility and treat that automated upgrade with more caution.

@Nuru to address your comment re: "which may not be for a long time" -- I see no reason why we should be wary of bumping major versions as long as we're correctly following semver and documenting the upgrade path. In this example of providing a default that will ensure future consumers are more secure, I believe that's justified. People don't read documentation a lot of the time, so I worry that if this is just a documentation change then it would lost to the majority. If we can make consumer environments more secure by default then we decrease future work for them once the infrastructure is already deployed in production. I see forward thinking security as one of the primary sources of value that we can provide to the community of users that consume these modules.

osterman:
> I wonder if we should release a v1.0 as it stands today without this change and then follow up quickly with a v2.0 that includes this change

Yes, I'm okay with either approach. But I'm not okay with saying we can't merge this change because it's changing the default (configurable) behavior and because we don't want to do an major release for such an insignificant change. 

Merging now, is fine for v0 semver. 
Cutting a release now for v1 fast followed by v2 is fine, if this is really as serious as you consider it to be. 

Doing nothing now seems like our process is too arbitrary to enforce. I don't want that.

oycyc (approved):

PR #300, opened 2024-03-12
Do not create ACL resource if Bucket Owner Enforced

### LINKED ISSUE #299: ACL fails if `s3_object_ownership` = `BucketOwnerEnforced`
### Describe the Bug

If we set the `s3_object_ownership` flag to `BucketOwnerEnforced`, the module fails with:

```
ACL: operation error S3: PutBucketAcl, https response error StatusCode: 400
```

### Expected Behavior

It should be possible to disable ACL creation, or it should be smarter (e.g. if BucketOwnerEnforced is set, don't create the ACL)

### Steps to Reproduce

Set `s3_object_ownership` to `BucketOwnerEnforced`

### Screenshots

_No response_

### Environment

_No response_

### Additional Context

_No response_

apankratiev-syberry:
## what

Do not create acl resource when s3_object_ownership=BucketOwnerEnforced

## why

ACL is not allowed when object ownership is BucketOwnerEnforced

https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html

## references

closes #299

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -6,6 +6,7 @@ locals {
   website_password_enabled  = local.website_enabled && var.s3_website_password_enabled
   s3_origin_enabled         = local.enabled && !var.website_enabled
   create_s3_origin_bucket   = local.enabled && var.origin_bucket == null
+  create_s3_acl             = local.create_s3_origin_bucket && var.s3_object_ownership != "BucketOwnerEnforced"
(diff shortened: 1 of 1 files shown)


joe-niland:
@Mergifyio refresh

apankratiev-syberry:
The bug is already fixed within #301

PR #301, opened 2024-03-12
BucketOwnerEnforced s3 buckets can't have an acl

mfuhrmeisterDM:
## what

disable creating acl resource in that case. 

## why

BucketOwnerEnforeced s3 bucket can't have an acl.

## references

* https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html


Benbentwo (approved):
LGTM

osterman (dismissed):
This is the wrong mergify config

Benbentwo (approved):

PR #330, opened 2025-01-11
memory and timeout vars for lambda@edge

mihaiplesa:
## what

Allow to configure memory size and timeout for Lambda@Edge module.

## why

These fields are not configurable now.

## references

Resolves https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/issues/331


mihaiplesa:
@coderabbitai full review

modules/lambda@edge/variables.tf
```diff
@@ -29,6 +29,8 @@ variable "functions" {
     source_zip   = optional(string)
     runtime      = string
     handler      = string
+    memory_size  = optional(number)
```
joe-niland, line 32: @mihaiplesa it could be handy to add a link to https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-at-edge-function-restrictions.html somewhere here to note that there are  specific Lambda@Edge limitations.

mihaiplesa: Added.

modules/lambda@edge/main.tf
```diff
@@ -78,6 +78,8 @@ resource "aws_lambda_function" "default" {
   function_name    = module.function_label[each.key].id
   runtime          = each.value.runtime
   handler          = each.value.handler
+  memory_size      = each.value.memory_size != null ? each.value.memory_size : 128
```
joe-niland, line 81: @mihaiplesa I just noticed these lines. Wouldn't it be better to set the default in the variables? i.e. `optional(number, 128)`. Lambda requires these values to be set so defaulting them should be fine for existing module users.

GabisCampana:
@goruha

mihaiplesa:
@joe-niland any idea why test are failing?

joe-niland:
@mihaiplesa I think it's due to SCP changes in the test account. I'll see if I can get an update.

GabisCampana:
@goruha

mihaiplesa:
Looks like tests finally passed, thanks @goruha.

goruha:
@mihaiplesa yes. But pls do not merge it yet. 
I'm still working on SCP configs

mihaiplesa:
@hans-d you're set as code owner on this, can you help by approving? Thank you in advance.

gberenice (approved):

PR #332, opened 2025-02-24
fix: Make sure tags are associated with Lambda functions

jwadolowski:
## what

All the tags provided by `cloudposse/label/null` should be associated with Lambda@Edge functions.

## why

[`tags`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function#tags-1) argument is not set at all which results in an empty tag list.

## references

https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/blob/v0.96.1/modules/lambda%40edge/main.tf#L78-L86


jwadolowski:
@RoseSecurity yup, the MR closure was a bit unexpected ending here 😉  As far as I can see the [`main`](https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/commits/main/) branch contains my commit, so it's all good 👍

PR #333, opened 2025-02-24
feat: Add support for custom Lambda@Edge policies

### LINKED ISSUE #261: Ability to attach a custom iAM policy to lambdas
### Describe the Feature

Custom lambda@edge permissions

### Expected Behavior

I wanted to provide a custom IAM policy but it doesn't seem possible since it's only attached at creation time I think.

### Use Case

Trying to permit access to other services for use within the lambda (e.g. S3), but can't.

### Describe Ideal Solution

The module would permit passing in a custom IAM role/policy to be attached.

### Alternatives Considered

Not using the module

### Additional Context

_No response_

jwadolowski:
## what

[Execution role](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function#role-1) associated with Lambda@Edge comes with a hardcoded policy that enables write access to CloudWatch logs. This PR adds support for additional policies. It was implemented in a similar fashion to [`additional_bucket_policy`](https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn?tab=readme-ov-file#input_additional_bucket_policy) from the parent module.

## why

It's a fairly common situation that a Lambda@Edge function needs access to other AWS services/resources than CloudWatch logs. `aws_lambda_function`'s [`role`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function#role-1) argument expects a single role ARN, therefore the only reasonable option is to append new policy statements to the IAM role created in scope of this module.

## references

closes #261


mihaiplesa:
@goruha @Gowiem @jamengual can someone start the tests please? Thanks in advance.

mihaiplesa:
Tests passed. Can we get an approval please @goruha @Gowiem @jamengual?

Gowiem (approved):
LGTM

jwadolowski:
An interesting finding that surfaced after this release: https://github.com/hashicorp/terraform/issues/36674

The code here and the module release procedure are perfectly fine. It seems to be caused by a regression bug in Terraform `1.11.1`.

PR #334, opened 2025-03-03
feat: Add environment variable support to Lambda@Edge functions

jwadolowski:
## what

Allow configuration of environment variables for Lambda@Edge functions via the [`environment`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function#environment-1) argument.

## why

Currently, the only option to provide dynamic configuration elements to a Lambda@Edge function is to use `templatefile()`. Using environment variables would be much cleaner and more maintainable.

While the `templatefile()` approach works, it becomes inconvenient when your function codebase is spread across multiple files and/or includes `node_modules`. This is because the entire codebase directory cannot be considered "static" and simply assigned to the function's `source_dir` argument. Instead, the templating process must occur before the directory is considered ready for zipping.


jwadolowski:
It turned out that Lambda@Edge does not support env variables ([docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-at-edge-function-restrictions.html)), which makes this PR pointless.

PR #344, opened 2025-06-12
feat(origin_groups): added support for custom group_id

### LINKED ISSUE #339: Allow configuration of origin_group name
### Describe the Feature

Currently, `origin_groups` is an object with following structure:
```hcl
list(object({
    primary_origin_id  = string
    failover_origin_id = string
    failover_criteria  = list(string)
  }))
```

It does not allow setting up custom id. 
Instead, the module names groups in this format:
`$name-group[$id]`

This is not ideal, as when specifying `target_origin_id` in `ordered_cache` config, we have to reference it by an arbitrary group id, based on its order of appearance in config, instead of actual name and purpose.

### Expected Behavior

Allow optional attribute to specify group_id:
```hcl
list(object({
    group_id = optional(string)
    primary_origin_id  = string
    failover_origin_id = string
    failover_criteria  = list(string)
  }))
```

### Use Case

Ability to specify custom `origin_group` id will allow creation of more readable, template-able and DRY tf manifests.

### Describe Ideal Solution

Allow optional attribute to specify group_id:
```hcl
list(object({
    group_id = optional(string)
    primary_origin_id  = string
    failover_origin_id = string
    failover_criteria  = list(string)
  }))
```

### Alternatives Considered

_No response_

### Additional Context

_No response_

atimofeev:
## what

Added support for optional `group_id` in `origin_groups` object, which can be referenced in `ordered_cache` block.

## why

The module code does not allow setting up custom id for origin group.
Instead, the module names groups in this format:
`$name-group[$id]`

This is not ideal, as when specifying `target_origin_id` in `ordered_cache` config, we have to reference it by an arbitrary group id, based on its order of appearance in config, instead of actual name and purpose.

## references

closes #339


atimofeev:
Any chance of getting this merged?

PR #358, opened 2025-09-01
fix: Resolve unsupported attribute error in S3 website block

### LINKED ISSUE #354: This object does not have an attribute named "redirect_all_requests_to"
### Describe the Bug

After an upgrade from 0.98 to 1.0, TF throws an error if a distribution has website enabled AND not specified redirect_all_requests_to . That's related to the website_config object, which doesn't have that field defined for the default

### Expected Behavior

Pass the build

### Steps to Reproduce

module "cdn" {
  version = "~> 1.0.0"

...

  website_enabled          = true
}

### Screenshots

_No response_

### Environment

_No response_

### Additional Context

_No response_

jwadolowski:
## what

Restore `lookup()` calls in `main.tf` to address the `website_enabled = true` use case that was broken when #340 replaced them with explicit variable calls to avoid silent default value assignments. Additionally includes corresponding module instances in the test suite.

## why

`website_enabled = true` implies a reference to 2 mutually exclusive configurations defined as the `local.website_config` variable. In the default case, `index_document`, `error_document`, and `routing_rules` elements exist, but `redirect_all_requests_to` does not, which leads to the following error:

```
╷
│ Error: Unsupported attribute
│
│   on ../../main.tf line 325, in resource "aws_s3_bucket" "origin":
│  325:       redirect_all_requests_to = website.value.redirect_all_requests_to
│     ├────────────────
│     │ website.value is object with 3 attributes
│
│ This object does not have an attribute named "redirect_all_requests_to".
╵
```

Similarly, `website_enabled = true` combined with `redirect_all_requests_to = "https://example.com"` would result in missing references to the `index_document`, `error_document`, and `routing_rules` fields.

All in all, in this particular case, the `lookup()` usage is definitely justified and does not mean a silent/hidden injection of a variable default.

## references

- resolves #354

### DIFF
--- a/examples/complete/website.tf
+++ b/examples/complete/website.tf
@@ -1,0 +1,34 @@
+module "website_default" {
+  source = "../../"
+
+  namespace = var.namespace
+  stage     = var.stage
+  name      = var.name
+
+  // Distinguish this module instance from the one in main.tf and to prevent S3 bucket name collisions
+  attributes = concat(var.attributes, ["website"])
+
+  cloudfront_access_logging_enabled = false
+
+  website_enabled = true
+
+  context = module.this.context
+}
+
+module "website_redirect_all" {
+  source = "../../"
+
+  namespace = var.namespace
+  stage     = var.stage
+  name      = var.name
+
+  // Distinguish this module instance from the one in main.tf and to prevent S3 bucket name collisions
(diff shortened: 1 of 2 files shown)


jwadolowski:
@hans-d @kevcube could you start the tests please?

milldr (approved):

PR #364, opened 2025-12-20
fix: Always ignore CloudFront distribution etag to suppress deployment-induced drift

npwolf:
## what

- Add `lifecycle { ignore_changes = [etag] }` to `aws_cloudfront_distribution` to suppress etag drift from Terraform plans
- No new variable — unconditionally ignores etag (see **why** below)

### The problem

CloudFront's `etag` is a read-only version identifier generated by AWS. It updates automatically whenever the distribution configuration changes — including as a side effect of **deploying Lambda@Edge or CloudFront Functions**. When a new function version is associated with the distribution, the etag changes even though no Terraform-managed attribute was modified.

This causes Terraform to surface a refresh-only diff on every edge function deployment. Since `etag` is computed and read-only, Terraform cannot revert it — the drift simply reappears on every plan until `terraform apply` is run to sync state.

### Why unconditional `ignore_changes = [etag]` is the right fix

The previous version of this PR used a variable to gate the behavior:

```hcl
ignore_changes = var.cloudfront_ignore_etag_changes ? [etag] : []
```

This is **invalid Terraform** — `lifecycle` blocks are evaluated at parse time and cannot reference variables or use expressions. The error is:

> `A static list expression is required.`

There is no valid way to make `ignore_changes` conditional on a variable. The only options are to always ignore `etag` or never ignore it.

**Always ignoring `etag` is safe** because:

- `etag` is not an input — it is a version identifier AWS generates. Terraform never uses it to configure the distribution.
- Terraform still uses the etag internally during `apply` for [optimistic locking](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_UpdateDistribution.html) to prevent concurrent modifications. This happens under the hood regardless of `ignore_changes`.
- Ignoring it makes plans actionable — you only see changes to attributes you actually defined in code.

### The one tradeoff: visibility of manual changes

If someone makes a manual change in the AWS Console, the etag will update. Without `ignore_changes`, Terraform surfaces this as drift (the "smoke detector"). With `ignore_changes = [etag]`, Terraform will not flag the etag change — but **it will still flag any actual attribute differences** (origins, TTLs, WAF associations, etc.) caused by that manual change.

The risk is only if someone makes a manual change that Terraform represents with zero visible attribute differences (pure metadata). In practice this is rare, and the tradeoff of eliminating persistent deployment-induced noise is worth it for teams using Lambda@Edge or CloudFront Functions.

## references

- Cloudfront etag and optimistic locking: https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_UpdateDistribution.html
- Terraform `lifecycle` static expression requirement: https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes


goruha:
@npwolf have you tested this fix on your side?

Lint failed with 

```
│ Error: Invalid expression
│ 
│   on ../../main.tf line 480, in resource "aws_cloudfront_distribution" "default":
│  480:     ignore_changes = var.cloudfront_ignore_etag_changes ? [etag] : []
│ 
│ A static list expression is required.
╵
```

goruha (changes requested):
Fix 

```
│ Error: Invalid expression
│ 
│   on ../../main.tf line 480, in resource "aws_cloudfront_distribution" "default":
│  480:     ignore_changes = var.cloudfront_ignore_etag_changes ? [etag] : []
│ 
│ A static list expression is required.
╵
```

npwolf:
Updated the approach based on @goruha 's feedback.

**Root cause of the error:** Terraform's `lifecycle` block is evaluated at parse/schema time, before variables are resolved. As a result, `ignore_changes` requires a static list of attribute references — it cannot evaluate variable references, function calls, or ternary expressions. The previous implementation:

```hcl
ignore_changes = var.cloudfront_ignore_etag_changes ? [etag] : []
```

will always fail with: *"A static list expression is required."*

**Fix:** Removed the `cloudfront_ignore_etag_changes` variable and made `ignore_changes = [etag]` unconditional. This is safe because `etag` is a read-only computed attribute — Terraform cannot modify it. The drift only occurs when an external process (e.g., deploying a Lambda@Edge function) updates the CloudFront distribution, and Terraform has no way to revert it regardless. Always ignoring `etag` prevents spurious plan noise without any risk.

